PR #275, opened 2015-05-16
Correct translation of ordinary of prime

fiapps:
It was pointed out to me that the English translation of prime is translating "opera tua" as "their works." I'm not sure if you have some reason for preferring that translation, but I've corrected it.


gregordick:
Thanks, Father. Merged.

PR #276, opened 2015-05-16
Typos and UTF-8 recognition problem

fiapps:
Most of these are typos, and uncontroversial. One could differ about the following changes:

* Regarding Sancti/08-30.txt, one could phrase the title in different ways, but since there's only one companion martyr, the plural was definitely wrong. The companion's name is unknown, and so he is called in Latin "Adauctus," the one who was added.
* The change to Sancti/12-26.txt removes a [BOM](http://en.wikipedia.org/wiki/Byte_order_mark) from the beginning of the file. Without this, [Rank] was unrecognized. Since 4889a8a, this may no longer be necessary, but it was already part of my commit.
* I more recently did the opposite to Santi/05-11.txt, adding a BOM to the beginning: without the BOM this file is not recognized as UTF-8, and so the word "fœnum" in lectio 2 is garbled (see [here](http://divinumofficium.com/cgi-bin/horas/Pofficium.pl?date1=05-11-2015&command=prayMatutinum&version=Rubrics%201960&testmode=regular&lang2=English&votive=)). It might be better to fix the UTF-8 recognition code rather than the case that's making it fail.

### DIFF
--- a/web/www/horas/English/Sancti/08-30.txt
+++ b/web/www/horas/English/Sancti/08-30.txt
@@ -17,7 +17,7 @@ smell of her perfumes, worthily themselves to become a sweet savour unto Christ.
 $Qui tecum
 
 [Commemoratio]
-!Commemoration of Ss. Felix and Companions Martyrs
+!Commemoration of St Felix and his Companion Martyrs
 @Commune/C3:Oratio proper
 $Oremus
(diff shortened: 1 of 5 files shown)


gregordick:
Thanks -- merged.

PR #279, opened 2015-06-25
Fix typos in 03-24 and 05-19 Latin, and 02-05 English

fiapps:
Correct typos on the feasts of St. Agatha and St. Gabriel (this latter text is used with the Divino afflatu rubrics), and in the commemoration of St. Pudentiana.


gregordick:
Thanks, Father. Merged.

Issue #280, opened 2015-07-06
Martyrology updates (1962)


rarty:
This is a great tool, and it includes probably the most complete/updated Martyrology available outside of a rare-books shop. 

I have a substantial update I'd like to make to the Latin Martyrology texts that would add accute accent marks and the common vowel ligatures. I checked against the latest Roman edition (ed. IV, 1956), and incorporated the changes from the 1961 "Variationes in Mart. Rom." As I'm not sure which editions the older historical versions of the office do/should use, I can't do much yet for those.

A few things I'd like other opinions on before I make a PR:

1. It looks like the "rubrics 1960" office currently uses the texts in the `Martyrologium2` folder, but since "reduced 1955" does too, I think that 1960 version should be distinct, i.e, `Martyrologium3`, and 1960-only changes removed from `Martyrologium2`.

2. Should the consonant `j` be retained *in toto,* or only for foreign/non-Latin names? (As in the modern Martyrology)

3. Can all the files be encoded UTF-8?

4. I can't find a rubric that prescribes the announcing of the year along with the moon each day, eg. "Anno Domini 2015" in the 1961 Breviary or in the 1956 Martyrology. Is this a received usage/monastic tradition?

Thanks,

KTL

ajmalton:
1 I agree, although if the differences are very few, I wonder if Laszlo had a good reason not to use just one set. How large is the difference?2 Unless the logic has been changed in the divinum code, then the rule is that j is replaced with i everywhere, starting with 1955. I don't think there's a way to retain J at all in the more modern rubrics settings (but I could be wrong, please someone correct me)3 ‎divinum data files can be in utf8 or latin1, with utf8 preferred. We really should switch them all to utf8 and get rid of the latin1 support, but work...4 it might be a Laszlo ‎Kiss tradition, like the usage of crosses etc which is a little quirkyAJM                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        From: K. T. LartigueSent: Monday, July 6, 2015 7:04 PMTo: DivinumOfficium/divinum-officiumReply To: DivinumOfficium/divinum-officiumSubject: [divinum-officium] Martyrology updates (1962) (#280)This is a great tool, and it includes probably the most complete/updated Martyrology available outside of a rare-books shop. 

[...]

PR #281, opened 2015-07-08
Updated text for Latin Martyrology

rarty:
As discussed in #280, this updates the text in `Martyrologium` with accents and ligatures, and other various differences according to the 1956 Martyrologium Romanum (ed. IV).

(Not implemented in the perl code yet) In `Martyrologium1955R` that text is modified to reflect the liturgical changes from 1955 through 1959 (28 affected days).

(Not implemented in the perl code yet) In `Martyrologium1961` that text is modified further to be in accord with the 1960 document *Variationes in Martyrologio Romano* and the 1962 rubrics (24 affected days).

I can alternatively make a unified version with conditional lines based on the office version, but would need to know what syntax to use...I can understand `(sed rubrica 1960)` before each modified line, but what about deletions?


gregordick:
Merged. Many thanks. I'll open another bug for the remaining work.

PR #289, opened 2015-08-05
Recent corrections

fiapps:
These are almost all corrections to typos. One corrects a mistranslation of "De vite mystica." The only one that could be questioned is adding a BOM to Pent03-1Feria.txt: without this, the file is read as CP-1252, which mangles the word "Israel". Arguably, the code to identify UTF-8 files should be improved instead of patching a single file.


gregordick:
Thank you, Father. Merged.

PR #294, opened 2015-08-25
Polish psalms & canticles

DLakomy:
I provided Polish version of psalms, canticles, Athanasian Creed and Gloria Patri... . I've added ”Polski” entry in the languages' list. Sources are given in credits.html.


gregordick:
Merged.  Many thanks.

PR #304, opened 2015-11-24
Next batch: preparation for comming days

mbab:
+AMDG+


gregordick:
Merged.  Many thanks.

PR #322, opened 2015-12-01
Missing Responsory Quad6-0

Lactantius:
Responsory 1 repeated as responsory 2. Replaced with correct responsory from English Breviary. This problem also found in the other versions of this file. 6-0r also has Jer. 2:18 as the last verse of lesson 1 and the first of lesson 2; the other two versions do not. Unsure if this is proper.


petermottola:
I checked my 1963 Collegeville: first lesson is 12-17 only, 18 begins the 2nd.

gregordick:
Thanks.  I've removed the extra verse from the end of Lectio1.

PR #329, opened 2015-12-03
Fix verse numbers

peterdiekertz:


gregordick:
The verse numbers in the psalms are broken in lots of places, but fixing this is complicated by the fact that the numbering is also used to split the psalms into their divisions.  In this case, we would also need to update `web/www/horas/*/Psalterium/Psalmi major.txt` to correct the reference to "142(9-12)" for Monastic Lauds, and fix the numbering in the psalm itself in all of the other languages.

There is also the question of exactly what the correct numbering should be, the solution to which is not obvious.

For the time being, I suggest that we leave the numbering as it is.  I don't mind merging this patch, though, if you would like, but if so, then please first update the translations of the psalm and `Psalmi major.txt`.

PR #339, opened 2015-12-09
Introduce cross-reference to C6a wherever possible

peterdiekertz:
(and typo in another file)


web/www/horas/Latin/Commune/C6a.txt
```diff
@@ -6,16 +6,10 @@ Psalmi Dominica
 Antiphonas horas
 
 [Ant Vespera]
-Haec est Virgo sapiens, * et una de numero prudentum.;;109
-Haec est Virgo sapiens, * quam Dominus vigilantem invenit.;;112
-Haec est quae nescivit * torum in delicto: habebit fructum in respectione animarum sanctarum.;;121
-Veni electa mea, * et ponam in te thronum meum, alleluia.;;126
-Ista est speciosa * inter filias Jerusalem.;;147
+@Commune/C6:Ant Vespera
```
gregordick: FYI: when the name of the section being reference is the same as the section doing the referencing, it can be omitted; so here you could just have said `@Commune/C6`.  Moreover, when one file is largely the same as another, you can put an `@`-directive at the start of the file, which will pull in *all* sections apart from those that are overridden, and then you can omit the sections that are the same as in the referenced file.  I'm happy for this to be merged as-is, though.

PR #344, opened 2015-12-11
Adding status page for Polish translation

mmolenda:
I'd like to add a simple page where we could track progress of Polish translation and present it in a concise way to Polish users in one central place. As an audience is fairly limited, I didn't link to this page from any other DivinumOfficium page; it's intended to be accessed through direct link.


gregordick:
I would prefer this to live in a page written in Markdown on the `wiki` branch, the link to which you could then pass on to interested parties.  I think this would be cleaner than having HTML generated from Markdown in source control.  Do you have a particular reason for wishing to have the file hosted on divinumofficium.com rather than on the wiki here?

mmolenda:
Page in the wiki branch works for me. I'll create another PR later.

PR #347, opened 2015-12-12
Adding status page for Polish translation

mmolenda:
A simple page where we could track progress of Polish translation and present it in a concise way to Polish users in one central place.


gregordick:
Merged.  Thanks for redoing this as a wiki page: I think it works much better this way.

PR #348, opened 2015-12-13
German updates

peterdiekertz:


peterdiekertz:
I cannot figure out why it says there are conflicts that must be resolved. AFAICS, only Polish texts were added into the main code since my last commit/pull request/whatever.
Will try it in a different way.

peterdiekertz:
Problem solved; it was a change which happened in Adv3-0 (latin) which I also edited. No idea why Github could not simply tell me that instead of referring to some weird command line...

Issue #355, opened 2015-12-17
Credo caption is displayed incorrectly in Mass - Propers mode


mmolenda:
When Propers mode is enabled on the Mass page (Propers link in bottom right), there's "Credo" caption with no content displayed in Latin column, which is expected, but on the translation side the caption itself consists of the whole prayer. 

I didn't go deep into the code, but it seems the problem occurs when the caption's code is exactly the same as its actual content.

The quick fix consists of adding Credo caption's translation to `www/missa/*/Ordo/Prayers.txt` file in every language, for example:

```
[Credo]
Kredo
```

If for certain translation this caption should sound "Credo", an HTML whitespace can be added to make the content different from caption's code:

```
[Credo]
Credo&nbsp;
```

PR #356, opened 2015-12-17
#355 Credo caption is displayed incorrectly in Mass - Propers mode

mmolenda:
Fix proposal for #355


gregordick:
I think it would be better to fix the bug rather than to work around it in the data files.  I'll work on a fix when I get the chance.

Issue #206, opened 2015-04-06
Cannot Generate Kindle-Compatible Version of Officium Defunctorum


GoogleCodeExporter:
```
What steps will reproduce the problem?
Using the settings that can generate Kindle-compatible file with the computer 
programme, the same file cannot be generated for Officium Defunctourm.

What is the expected output? What do you see instead?
Expected: Table removed. Instead, the table is still there.

What version of the product are you using? On what operating system?
Most updated version from the website. I am using Windows 7.

Which rubric (19600, 1955, etc.) are you using?
1960

Please provide any additional information below.


```

Original issue reported on code.google.com by `hk...@connect.hku.hk` on 15 Feb 2013 at 5:02

gregordick:
The old standalone program is now deprecated.  Suitable EPUB files can be generated using one of the tools in `tools/standalone/`.

Issue #36, opened 2015-04-06
Downloadable files


GoogleCodeExporter:
```
The downloadable files on 
http://www.divinumofficium.com/www/horas/Help/download.html should be updated 
regularly with the latest edits that we have made to the repo.

We also need to add a link on this page, and preferably on the home page as 
well, to the Google Code site.

On the download page, the wording could be:

The repository for the code for this site can be found at 
http://code.google.com/p/divinum-officium.  This site provides a Subversion 
repository for the Divinum Officium Project, so that edits can be made to the 
code while they are made accessible to the entire team.  Errors may also be 
posted as "Issues", located on the heading of the Google Code site. This 
repository always has the most up-to-date version of the files, as the 
repository goes live from here to the website usually every Saturday.

Any corrections or suggestions?
```

Original issue reported on code.google.com by `canon.mi...@gmail.com` on 7 Oct 2011 at 4:17

PR #358, opened 2015-12-19
An alternative EPUB generator.

imjagpul:
This script generates the office texts in EPUB format for one or several years.
Resulting files can easily be converted to MOBI format using kindlegen.

It does not rely on ebook-convert to do the conversions, but the XHTML files
(necessary for the EPUB format) are generated using customized perl scripts.
Consequently, the resulting files are considerably smaller and the production
is faster.

Up-to-date generated files using 1962 rubrics are maintained on:
http://breviarium.srubarovi.cz

Relevant for issues #36 and #206.


gregordick:
Merged.  Thank you very much!

Useful future work would be to factor out code that is shared with the main Divinum Officium implementation.  There's gradual work on the `calcalc` branch to pull things out into modules that should make this easier.

PR #361, opened 2015-12-21
Polish translations for the 24th and the 25th of December

DLakomy:
Everything should be in Polish in these days (Rubrics 1960).


gregordick:
I'll update the live site.

PR #363, opened 2015-12-25
Polish - Mass - adding Sanctissimi Nominis Jesu, Sanctae Familiae, Epiphany 01.06, Baptismatis 01.13

mmolenda:


gregordick:
Merged.  Many thanks.  Two comments:

1. Where a file in a translation is identical to the Latin, that file may (and ideally should) be omitted altogether.
1. Where possible, please avoid including merge-from-upstream commits in pull requests.

mmolenda:
Understood, I'll comply to these rules in future commits.

PR #366, opened 2015-12-26
Add links to Wayback Machine (Missal sources)

lemtom:
A few of these Missal sources are no longer online. I have added links to the archived webpages on the Wayback Machine. The design of the links have been slightly changed to accommodate for this.


gregordick:
Merged.  Many thanks.

Issue #373, opened 2016-01-02
St. Telesphorus Oration missing


peterdiekertz:
5 January, Trient 1570 and 1910 options.
"Commune/C2:Oratio1 Gregem missing" is shown instead of the oration,
even though
1. the oration Gregem is in fact in the files
2. a different one should be shown, as the Common of Popes was only introduced under Pius XII.

mbab:
Line 42 in file 01-05.txt should look like

`@Commune/C2:Oratio1:s/N\./Telesphori/`

gregordick:
The proposed solution would result in the wrong prayer for Divino and later rubrics, where it should be 'Gregem tuum'.  The 'Gregem' keyword in cross-references is handled specially.  This probably needs a code change to fix.

mbab:
I'm sorry.
Looking at code the regexp in file specials.pl line 1818 is probably broken
it gives in example: `$item == "Oratio1 Gregem"` and empty  `$after` variable instead of
`$item == Oratio1` and `$after == Gregem`, 
This needs to change crossref rules for Commoratio.

peterdiekertz:
I see it works correctly for e.g. St. Linus (23 Sept). In his case, the "old" oration is the same one as for St. Telesphorus (from the Common of Martyr-Pontiffs, 2nd place), though in Sancti/09-23.txt the oration is copied, not cross-referenced. If that is a workaround, we probably don't want to use it.

PR #374, opened 2016-01-02
Correction and issue #373 patch

mbab:


gregordick:
The fixes for #373 are not quite right: please see there for discussion.  I'm happy with the Polish changes, though.

mbab:
Replaced with my next PR

PR #378, opened 2016-01-03
Make 01-01o and 02-22t use the "normal" file for nearly everything + other cross-refs

peterdiekertz:


mbab:
English version of file 02-22t.txt also should be removed. If exist it breaks translations.

PR #388, opened 2016-01-07
Mass - Adds a French translation of the Ordo

LGD-Fr:
Why "French" and not "français" ? Because it did not work with "français"… Maybe the "ç" ?


gregordick:
Thanks very much: looks excellent.  Before I merge, could you please add any sources used for the translation to `web/www/horas/Help/credits.html`?

UTF-8 in paths is probably not handled correctly, hence the problem with 'français'.  I'll open a bug to track the implementation of a fix for this.

LGD-Fr:
I also must check the permissions and intellectual properties, I pulled request too early.

LGD-Fr:
The credits are now updated @gregordick ; we also added some translations of the propers.

gregordick:
Merged.  Thanks very much!

PR #392, opened 2016-01-09
Fixes and Polish translations

mbab:
+AMDG+

### DIFF
--- a/web/cgi-bin/missa/propers.pl
+++ b/web/cgi-bin/missa/propers.pl
@@ -82,7 +82,9 @@ sub specials {
     if ($item =~ $section_regex) {
       $label = $1;
 
-      if ($rule =~ /omit.*\b$label\b/i) {
+      if ( ($rule =~ /omit.*\b$label\b/i) ||
+           ( ( $version =~ /1570/) && ( $item =~ / Le/ ) ) # omit Leonine prayers issue #367
+      ) {
         # Skip omitted section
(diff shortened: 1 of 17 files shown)


web/cgi-bin/missa/propers.pl
```diff
@@ -82,7 +82,9 @@ sub specials {
     if ($item =~ $section_regex) {
       $label = $1;
 
-      if ($rule =~ /omit.*\b$label\b/i) {
+      if ( ($rule =~ /omit.*\b$label\b/i) ||
+           ( ( $version =~ /1570/) && ( $item =~ / Le/ ) ) # omit Leonine prayers issue #367
```
gregordick: A couple of things:

 - Identifying the Leonine prayers by this regex is far from ideal.  Better would be to arrange for the title of the section to be the same in all translations, so that it can be identified robustly, and then translate it for display as is done with the other section headers.
 - There are superfluous parentheses and whitespace here.

I'll merge this change as it is for now, though, as the code is already full of such things, but it would be best in future to try to make things more robust.

gregordick:
Merged: thank you.  In future it would be helpful to put code changes in separate pull requests from data-file changes (and, ideally, to keep code changes fixing different bugs in different PRs).

web/cgi-bin/missa/propers.pl
mbab: 1. I know it is poor but title section in translations are out of order, all should be rewritten.
2. Will keep in  mind.

Issue #394, opened 2016-01-10
Palm Sunday Mass - some of 1955 changes to Passion are not taken into account  


mmolenda:
There were changes made by Pius XII in 1955 to the Holy Week. One of the changes was reduction of Passion reading in Palm Sunday propers – removing of first 35 verses and the last verse. Divinum Officium has two files for Palm Sunday, `Quad6-0.txt` and `Quad6-0r.txt`, that are supposed to reflect this change, though only removal of latest verse is reflected.
@gregordick proposed following solution:

A possible solution (to avoid duplicating texts) would be to do
something like this:

```
# In Quad6-0:
[Evangelium1]
#26:1 - 26:35 go here.

[Evangelium2]
#26:36 until just before 'Erat autem' go here.

[Evangelium3]
# 'Erat autem' to the end goes here.

[Evangelium]
@:Evangelium1
@:Evangelium2
@:Evangelium3

# In Quad6-0r:
[Evangelium]
@Tempora/Quad6-0:Evangelium2
```

PR #401, opened 2016-01-12
Polish - Mass - Adding Holy Monday, Holy Tuesday

mmolenda:
I found a problem in Holy Tuesday's Passion. It's similar to one described in issue #394. After 1955 changes, Passion starts with verse Marc 14:32, while Divinum Officum shows it from 14:1.
I made a fix to Latin and Polish propers. If this solution is accepted I can fix other languages.


gregordick:
Thanks, looks good.  I'll take your word that that use of conditionals works as intended!

PR #406, opened 2016-01-15
Issue404

mbab:
+AMDG+

### DIFF
--- a/web/cgi-bin/horas/horas.pl
+++ b/web/cgi-bin/horas/horas.pl
@@ -212,11 +212,7 @@ sub resolve_refs {
     elsif ($line =~ /^\s*\!\!(.*)/) {  
       my $l = $1;
       my $suffix = '';
-      if ($l =~ /\{.*?\}/) {
-        $l =~ s/(\{.*?\})//; 
-  	    $suffix = $1;
-		    $suffix = setfont($smallblack, $suffix); 
-      }
+      if ($l =~ s/\{[^:].*?\}//) {$suffix = setfont($smallblack, $&);}
       $line = setfont($largefont, $l) . " $suffix\n";   
 	  if ($expand =~ /skeleton/i) {$line .= linkcode1();}
     } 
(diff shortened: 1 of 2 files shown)


web/cgi-bin/horas/horas.pl
```diff
@@ -212,11 +212,7 @@ sub resolve_refs {
     elsif ($line =~ /^\s*\!\!(.*)/) {  
       my $l = $1;
       my $suffix = '';
-      if ($l =~ /\{.*?\}/) {
-        $l =~ s/(\{.*?\})//; 
-  	    $suffix = $1;
-		    $suffix = setfont($smallblack, $suffix); 
-      }
+      if ($l =~ s/\{[^:].*?\}//) {$suffix = setfont($smallblack, $&);}
```
gregordick: Please avoid `$&` (and ``$` `` and `$'`).  The server is running perl 5.10 and the use of those variables makes matches more expensive on old perls.  The generic way around this is to use an explicit capture group around the whole pattern.

web/cgi-bin/horas/specials.pl
```diff
@@ -16,8 +16,11 @@ sub specials {
   my $lang = shift;		 
   
   $octavam = ''; #check duplicate commemorations
-  my %w = (columnsel($lang)) ? %winner : %winner2; 
-  
+  my %w = (columnsel($lang)) ? %winner : %winner2;
+
+  my $hymntrans = $translate{$lang}{Hymnus};
```
gregordick: Maybe this should be

    my $hymntrans = 'Hymnus' || $translate{$lang}{Hymnus};

This is supposing that `$hymntrans` is used again later on (which I imagine that it is, given the last hunk of the patch).

mbab: This always gives `'Hymnus'`
If you mean
```
my $hymntrans = $translate{$lang}{Hymnus} || 'Hymnus';
```
it is not necessery as translate always contains default Latin values

gregordick: Good points!

PR #424, opened 2016-01-25
Update PlHoras.md

DLakomy:
The link to "Katedry Sw. Piotra, Apostola (przed 1960)*" appears to be broken. I'm unable to correct this at the moment. Probably I pass the parameter "version" in a wrong way.

I receive an error:
Can't use string ("") as a HASH ref while "strict refs"...


gregordick:
The problem is the quotes in `'Reduced 1955'`.  They should be omitted.

DLakomy:
It's working now. Thank you!

PR #430, opened 2016-01-30
Polish texts for 2nd February

DLakomy:
The PR contains a complete set of the texts needed on the Feast of the Purification.

I would greatly appreciate it if you could update the live site before this feast, so it could be used then.


gregordick:
Thanks.  Site updated.

DLakomy:
Thanks a lot!

PR #436, opened 2016-01-31
Latin - Mass - adds forgotten verses to a tractus

LGD-Fr:


gregordick:
This has been merged as part of #434, but that patch series also reverted the change.  Was this intentional?

PR #444, opened 2016-02-05
Translation of 'Commemoratio'

mbab:
+AMDG+

### DIFF
--- a/web/cgi-bin/horas/specials.pl
+++ b/web/cgi-bin/horas/specials.pl
@@ -1415,8 +1415,8 @@ sub getcommemoratio {
   postprocess_vr($v, $lang);
 
   our %prayers;
-  $translate{$lang}{Commemoratio}=~s/\n//;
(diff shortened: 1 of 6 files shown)


web/cgi-bin/horas/specials.pl
```diff
@@ -1415,7 +1415,8 @@ sub getcommemoratio {
   postprocess_vr($v, $lang);
 
   our %prayers;
-  my $w = "!Commemoratio $rank[0]\nAnt. $a\n_\n$v\n_\n$prayers{$lang}->{Oremus}\nv. $o\n"; 
+  $translate{$lang}{Commemoratio}=~s/\n//;
+  my $w = "!$translate{$lang}{Commemoratio} $rank[0]\nAnt. $a\n_\n$v\n_\n$prayers{$lang}->{Oremus}\nv. $o\n";
```
gregordick: - Please call `translate()` rather than looking up the hash directly.
- (I know this was already not the case, but:) Please wrap source lines at 80 characters.

PR #473, opened 2016-03-02
Update Quad3-3t.txt

bbloomf:
fixed typo "aureo s" => "aureos"


bbloomf:
I just noticed that in the Gospel, there is also the misspelling "Accessérurit" which should be "Accessérunt"

PR #474, opened 2016-03-04
Polish translations

mbab:
+AMDG+


web/www/horas/English/Sancti/03-19.txt
```diff
@@ -229,24 +229,9 @@ the custom of the law,
 V. And His father and mother marvelled at those things which were spoken of Him.
 R. Simeon took Him up in his arms, and blessed God.
 &Gloria
+A
```
gregordick: I don't think this line should be here.

mbab: Sure.

mbab:
Replaced with #492

PR #475, opened 2016-03-04
Update Quad3-5.txt

bbloomf:
In the gospel: removed errant text, fixed typo (ci => ei)


gregordick:
Thanks.  If possible, please either preserve the encoding of the files, or make the encoding changes in separate commits.  This will make the diffs easier to read.

PR #486, opened 2016-03-23
Reverting commit 5a9dc6b as it causes problems in Mass part

mmolenda:
Removing these files results in lack of Polish Mass translation for Maundy Thursday, Good Friday and Easter Vigil. Perhaps Polish files for the Mass can be adjusted to allow deletion of files in question but I'd defer it for time after Easter.

http://divinumofficium.com/cgi-bin/missa/missa.pl?lang2=Polski&Propers=1&date=3-24-2016
http://divinumofficium.com/cgi-bin/missa/missa.pl?lang2=Polski&Propers=1&date=3-25-2016
http://divinumofficium.com/cgi-bin/missa/missa.pl?lang2=Polski&Propers=1&date=3-26-2016

### DIFF
--- a/web/www/horas/Latin/Tabulae/Tr1960.txt
+++ b/web/www/horas/Latin/Tabulae/Tr1960.txt
@@ -1,6 +1,9 @@
 C03-20=Votive/Coronatio;;
 Tempora/Quad5-5=Tempora/Quad5-5Feria;;
 Tempora/Quad6-0=Tempora/Quad6-0r;;
+Tempora/Quad6-4=Tempora/Quad6-4r;;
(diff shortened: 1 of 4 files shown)


gregordick:
Merged.  @mbab: Could you please revisit your original change to find a suitable solution?

mbab:
I checked all Missa files now and see no other problems as apropriate files are also same.
The only difference are in Quad6-4t and Quad6-4 files, in later file there is 'noPax' rule, and  few differnce in comments and format, in the first file in 'Lotio pedum' there is no Evangelium specified only in comments. Please decide if there is need to revoke file Quad6-4t.

PR #487, opened 2016-03-25
Add missing Alleluia's to C7ap

DLakomy:
Correct me if I'm wrong, but I think this alleluia's should be there.


gregordick:
Merged.  At least for the antiphons and versicles, it was probably harmless that they were missing, as they should be added automatically.  Not sure about the responsories.  In any case, the change is an improvement -- thanks!

PR #502, opened 2016-04-11
Latin Tempore files: inner references

mbab:
+AMDG+


gregordick:
Thanks very much for this.  This sort of change is much needed and much appreciated!

PR #529, opened 2016-05-25
Polish translations

mbab:
+AMDG+


DLakomy:
I see that few responsories have been added in 681cb7a in Latin, as placeholders I guess. Wouldn't it be better to use the English ones? They would be at least partially understandable for Polish users.

PR #534, opened 2016-05-28
Change the preface on Festum Sanctissimi Corporis Christi

### LINKED ISSUE #533: Wrong preface on Festum Sanctissimi Corporis Christi (1960)
According to the rubric number 498, and to the propers given in the missal, it should be Communis, not de Nativitate. It can be checked here: http://musicasacra.com/2007/10/missale-romanum-1962/ (page 32 and 458, according to the PDF numbering).

I will provide a suggested fix in a moment. Please tell me, if it's ok.

DLakomy:
Fixes #533


DLakomy:
Merged.

PR #551, opened 2016-06-16
Fix Saturday pre-Pius

peterdiekertz:
Currently the Saturday Matins in our Trient modes are broken: Ps. 97 98 91 100 101 102 are shown, whereas it should be 97-108. Visible e.g. on 18 June and 9 July 2016.

I don't know if this fixes the whole problem (with only 6 psalms being shown, even though the code at the moment already provides for 12), but it will at least take out the wrong 91.


peterdiekertz:
Done.

Issue #552, opened 2016-06-17
Matins responsory in Paschal time


mbab:
which are taken from Commune can have automatically inserted Alleluia's (as Ant,Versum, etc).
This will allow to clean C[4-7,11]p files.

gregordick:
Merged in 76ccd7f.

PR #555, opened 2016-06-17
 #552 implementation

mbab:


gregordick:
Thanks!  I'm happy with the approach of the patch, although I haven't though it through in complete detail (for example whether those indices into `@resp` are always correct, but probably they are).

I'm working on a script that will generate diffs in outputs across revisions and pull requests.  Is it OK if I wait for this to be ready before merging this PR?  This will probably take a two or three weeks.  I can merge it sooner if it blocks your other work, though.

A few stylistic points.  No need to address them for this PR, but for future reference:

 * Please avoid prototypes on subroutines (i.e. things like `(\$$)`).  I used them for a while (largely because my background is in C), but in Perl they cause more trouble than they save.
 * Please use whitespace around `=` operators and after comma operators.
 * Please use [Allman-style braces](https://en.wikipedia.org/wiki/Indent_style#Allman_style) for new subroutines; otherwise, please match the style of the surrounding code.

gregordick:
Looked good with a manual run of the regression script.  Sorry that took me so long...

PR #565, opened 2016-07-04
Latin Sancti files: 08-* more

mbab:
+AMDG+


web/www/horas/Latin/Sancti/06-29.txt
```diff
@@ -16,7 +16,7 @@ Psalm5 Vespera3=138
 [Ant Vespera]
 Petrus et Joannes * ascendebant in templum ad horam orationis nonam.
 Argentum et aurum * non est mihi: quod autem habeo, hoc tibi do.
-Dixit Angelus ad Petrum: * Circumda tibi vestimentum tuum, et sequere me.
```
gregordick: I checked several 19th- and 20th-century breviaries and the 1912 Antiphonale, and they all placed the asterisk after "Petrum:".  The Liber usualis was the only exception.  I think we should keep the existing position.

Issue #244, opened 2015-04-06
1962 Italian Lauds 06-25: Commemoration of octave appears


GoogleCodeExporter:
```
http://divinumofficium.com/cgi-bin/horas/Pofficium.pl?date1=6-25-2014&command=pr
ayLaudes&version=Rubrics%201960&testmode=regular&lang2=Italiano&votive=

There's a Commemoratio 2 in 06-25, which is suppressed correctly in all 
languages other than Italian.
```

Original issue reported on code.google.com by `igregord` on 25 Jun 2014 at 9:47

Issue #263, opened 2015-04-06
Italian Lauds for 12/11/2014  has Commemoration of Octave of Immaculate Conception


GoogleCodeExporter:
```
What steps will reproduce the problem?

1. Select Lauds for 12/11/2014 
2. Select Italian
3. Use default 1960 Rubrics

What is the expected output? What do you see instead?

Italian has the commemoration for the Octave of the Immaculate Conception, 
which should not be there according to the Latin.



```

Original issue reported on code.google.com by `roberto....@gmail.com` on 11 Dec 2014 at 7:50

PR #579, opened 2016-08-02
Commemoratio fixes

mbab:
+AMDG+


mbab:
Also fixes: #244, #263.

PR #583, opened 2016-08-10
English files: conform previous latin changes

mbab:
+AMDG+


fiapps:
Please test these changes carefully with `admin/divinum-replay.pl` to make sure they do not introduce errors in the output. In particular, I see you've made the same change to Responsory4 in English/08-15 that you made in the Latin, resulting in the problem that #589 corrects. I would expect your change here to produce the same error in the English (a spurious asterisk).

I've caught some errors introduced by the replacement of texts with references, but my testing is not exhaustive, and other errors may go unnoticed.

mbab:
08-15: The english version had no second star, but capitalization was wrong so I correct them. There are also punctation differences but I have no english version matutinum sources to decide which one is correct.

fiapps:
The English translation was originally drawn from the translation of the Marquess of Bute, which was [printed in 1908 in four volumes](http://www.newliturgicalmovement.org/2006/12/internet-archive-roman-breviary.html). However, since then @APMarcello3 has made many changes to modernize the language.

gregordick:
I'll look at this one when my regression-test script is ready (which regrettably I haven't had time to work on recently).

Issue #277, opened 2015-06-15
Prime of the Little Office is missing Ps. 53


gregordick:
E.g. http://divinumofficium.com/cgi-bin/horas/Pofficium.pl?date1=6-15-2015&command=prayPrima&version=Rubrics%201960&testmode=regular&lang2=English&votive=C12.

Divino seems okay.

DLakomy:
Problem solved via #590.

Issue #506, opened 2016-04-16
Missing/wrong elements in the Little Office


gregordick:
In 1955/1960, Ps. 128 is missing: http://divinumofficium.com/cgi-bin/horas/Pofficium.pl?date1=4-16-2016&caller=&command=prayCompletorium&version=Rubrics%201960&testmode=regular&lang2=English&votive=C12

This is present in Divino and earlier.

With all rubrics, 'Converte nos', and the Gloria after 'Deus in adjutorium' are missing.

CC @DLakomy.

DLakomy:
'Converte nos', and the Gloria after 'Deus in adjutorium' are still missing.

gregordick:
> 'Converte nos', and the Gloria after 'Deus in adjutorium' are still missing (Compline).
> 
> One more observation: the verse 'Diffusa est gratia' is missing before 'Pater noster' in Matins.

@mbab: Could you please take a look at this one?

DLakomy:
I think the Antiphona ad BMV is missing in the Compline. I see it in my Diurnale, but not on the website.

DLakomy:
@igneus This issue concerns the Little Office (Officium Parvum).

igneus:
@DLakomy Oh, I'm sorry for the noise. I must have skipped the title when reading the comments.

DLakomy:
There is more issues:
- there is no Gloria Patri in [Responsory2] in C12A (it is correct), it is present on the website, though which is wrong),
- conclusion missing during the Advent (I haven't tested other seasons), all hours,
- lack of Gloria Patri and Alleluia at the beginning, in minor hours,
- wrong capitulum and antiphon during the Advent (I haven't tested other seasons); however, the file C12A seems to be ok.

mbab:
Close?

DLakomy:
Few things remain:

1. Wrong capitulum, verse and antiphon in Vesperæ (tested using 1960 rubrics, during Advent; I'm unsure about the other seasons).
2. Antiphon to BMV missing in Compline.

By the way: other languages have to be conformed, to include the corrections. I'm unsure if it's a reason to leave this issue open.

gregordick:
Let's leave it open until everything's fixed.

mbab:
Please let me know if #680, doesn't fix this or close this.

DLakomy:
[...]

PR #590, opened 2016-08-11
#506 fix

mbab:
Also needed data files changes with Triduum references, corrections (sp. in 1955/60) in all languages.

+AMDG+


mbab:
Also fix #277

PR #593, opened 2016-08-20
Polish translations - Sancti+Tempora: September, Tempora: Pent18-Pent24 + Commune

mbab:
+AMDG+

### DIFF
--- a/web/www/horas/Polski/Sancti/09-15.txt
+++ b/web/www/horas/Polski/Sancti/09-15.txt
@@ -104,12 +104,15 @@ Aby mi w trudach walki udzieliła mocy,
 Takiej, jaka pod Krzyżem była w duszy Marji.
 Amen.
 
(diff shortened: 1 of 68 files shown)


DLakomy:
I see two issues:
1. Few texts are in Latin (English is preferred for untranslated parts)  in: 9-15, 9-16, 9-17, 9-29 (I hope I listed all the files). They appear to be short, so I think I'll translate them, at least the ones not from Matins. I'll send these lines via email today.
2. The change in C1a (a link to C1) seems to not work properly as long as the English version isn't changed  analogically. Do you plan to change the English file soon? I tested this by checking the responsory [here](http://divinumofficium.com/cgi-bin/horas/officium.pl?language2=Polski&command=praySexta&browsertime=9-21-2016&version=Rubrics%201960) (September 21st, Sexta, 1960).

PR #598, opened 2016-08-25
English files: conform previous latin changes in C1a

mbab:
+AMDG+


DLakomy:
I've tested the changes locally and everything seems to work correctly. Can this PR be safely merged? Since the reference is already present in Polish C1a, the change in English version is necessary for this to work. If this PR remains unmerged, the Polish texts on September 21st won't be rendered correctly.

fiapps:
I've tested it as well, and found no problems.

PR #599, opened 2016-08-26
consistency in the name & cross-refs

peterdiekertz:


mbab:
BTW. What is purpose of files ended with `a`?
In `Tabulae` only `Epi1-0a` is used.
Or it is hardcoded somewhere?

gregordick:
I don't think there's any special handling for '`a`' files.  IIRC they're used for 1570 in cases where it differs from 1910 (or is it the other way around?!).  This particular file is cross-referenced by 08-25.

Issue #582, opened 2016-08-08
Abbreviations of books


luc--:
Hello,

I haven't found any tables for all the abbreviations used in DivinumOfficium (only this pagehttp://divinumofficium.com/www/horas/Help/Abbreviations1.html)… It seems that all abbreviations aren't the same for each translation or even not the same in a single abbreviation.

For example, the book of Jeremy is written `Jerem` or `Jer`, or `Ex`, `Exod` and `Exodus` for the book of Exodus, etc. We can also note that abbreviations are sometimes ended by a point.

As I'm actually working on an automatic conversion from French text files (provided by the Abbaye du Barroux) to the format used in DO, I have to do a list of the abbreviations used in DO. Wouldn’t it be an occasion to have the same abbreviations everywhere in any languages ?

I would suggest using abbreviations based on Latin names of the books of the Bible : Gen, Ex, Lev, Num, Deut, Ios, Iud, etc. It would then be simple to use them in any language of DO by a simple parsing script.

---

PS : A useful method to find abbreviations used in DO :

``` bash
grep -rnw <directory_of_DO> -e "\!abbreviation"
```

gregordick:
Standardising the abbreviations for each language would be an excellent idea.  I agree with your suggestions for the Latin; furthermore, I think it would be best to those used by a particular edition of the liturgical books where possible.

> It would then be simple to use them in any language of DO by a simple parsing script.

Could you please describe this suggestion in more detail?

fiapps:
> I think it would be best to those used by a particular edition of the liturgical books where possible.

I think the 1961 editio typica would be the edition to follow. This is the official reference for the version we call "Rubrics 1960." Its abbreviations are terminated with a period: Luc., Rom., 1 Tim., 1 Ioann., etc. Unlike DO, it does not use book name abbreviations in the readings of matins, where the name of the book has just been given in full. For example, DO today has (using italics for rubrics):

> **_Lectio 1_**
> Lectio sancti Evangelii secundum Matthaeum.
> _Matt 16:24-27_

The _editio typica_ has:

> Lectio sancti Evangelii secundum Matthaeum.
[...]

PR #613, opened 2016-09-19
Abbreviations of books 

luc--:
Hello !

This pull request changes all abbreviations of the Bible books used in both directories hora and missa.
A script is provided in the admin directory (with some POD format documentation).

(issue #582)


LGD-Fr:
@luc--,i seems your script missed the books with numbers (1 Reg, 2 Cor etc.)
I wrote a shell script to correct this. (20dc390)

LGD-Fr:
I fixed the bash script (5c9a4a2), so now latine abbreviations are used also for numbered books (1 Reg etc.)

fiapps:
Personally, I don't see why it is desirable to use the Latin abbreviations in other languages, which are offered as translations of the Latin. Why translate the text and not the abbreviation? Anyone who wants the Latin abbreviation can see it in the Latin column next to the translation. On the other hand those who are viewing a translation are presumably doing so because they are not fluent in Latin, and may therefore not recognize the Latin abbreviations.

Nevertheless, I leave it to @APMarcello3 to decide what to do for the English translation.

luc--:
Hello, I’ve made an answer in the issue #582.

PR #618, opened 2016-10-04
Latin files: Sancti/10-* and few other - references + corrections

mbab:
+AMDG+

PR #619, opened 2016-10-05
English files: conform previous latin changes

mbab:
+AMDG+


DLakomy:
Everything seems to be ok. I think it can be merged as soon as #618 is.

PR #623, opened 2016-10-31
Polish files: November

mbab:
+AMDG+


DLakomy:
Two minor issues:
1. Latin texts in Tempora/112-3.txt and Tempora/113-3.txt.
2. Two html tags (`</p>`) are present in Lectio2 section, in Tempora/113-3.txt.

Apart from that, everything seems to be ok.

PR #634, opened 2016-11-18
Advent Polish translation

mbab:
+AMDG+


DLakomy:
I've found two typos:
- Adv1-0:  „Adwewntu”,
- Typo in Adv3-6 „...Sychych Dni...”.

PR #636, opened 2016-11-18
Latin Sancti in December: corrections and improvements

mbab:
+AMDG+


web/www/horas/Latin/Sancti/12-27.txt
```diff
@@ -225,23 +233,23 @@ $Deo gratias
 Exiit * sermo inter fratres, quod discipulus ille non moritur; et non dixit Jesus, Non moritur: sed, Sic eum volo manere, donec veniam.
 
 [Octava 2]
-!Commemoratio octavae S. Joannis.
-Ant. Iste est Joannes, qui supra pectus Domini in cena recubuit: beatus Apostolus, cui revelata sunt secreta caelestia.
+!Commemoratio Octavae S. Joannis
+Ant. Iste est Joannes, qui supra pectus Domini in caena recubuit: beatus Apostolus, cui revelata sunt secreta coelestia.
```
peterdiekertz: Shouldn't this stay as "cena" or become "coena"?

rl1234567890: Accordign to Lewis & Short, it should be cena 
http://www.perseus.tufts.edu/hopper/text?doc=Perseus%3Atext%3A1999.04.0059%3Aentry%3Dcena
Also not only caelestia is perfectly fine but is is also preferable:
http://www.perseus.tufts.edu/hopper/text?doc=Perseus%3Atext%3A1999.04.0059%3Aentry%3Dcaelestis

fiapps: The Vatican latinists agreed: the 1961 _editio typica_ has _cena_ and _cælestia_ (27 Dec., Ant. ad Benedictus, p. 291a).

mbab: The only point can be, that this Ocatvae was abandoned, when writing (in 1902) as 'coena' and 'coelestia'.
Of course this provides some inconsistency, which can be? cured in code.

fiapps: I think it would be difficult to use different orthographic styles in different sections of the same file. At the very least, if this is to be done, it will be necessary to document the style to follow in each section.

However, if there is going to be code that changes the orthographic style, then there is no need to use different styles in different parts of the database of texts: all texts can be kept in a single style, and then converted to the style chosen for output.

Another consideration is that I doubt that 1902 edition was an *editio typica.* That means that there could have been other editions at the same time following different orthographic styles, based on the preferences of their publishers.

mbab: So, something should be done?
both words appears in `Ant 1`, (this is why I changed them in `Commemoratio`)
and `coelestia` is also in `Responsory8`

fiapps: At least as far as the 1960 rubrics are concerned, `Ant 2` should to be corrected to _cena_ and _cælestia,_ and also in `Responsory8` this spelling should be followed. The 1961 _editio typica_ is the official model in these cases. If `Ant 2` is going to continue to be `@:Ant 1`, then the change has to be made in `Ant 1`.

As for the Commemoratio, there isn't an official model I'm aware of, and I don't personally have any authority to make decisions for the Divinum Officium Project, but I think following the same style throughout the database of texts is the simplest practice.

gregordick:
LK followed the orthography of the books from which he was working (an 1888 breviary and the Sandhofe missal, I think).  We don't make any attempt to adjust this for different rubrics.  I'd suggest leaving things as they are as a general principle, but fixing up inconsistencies within individual files, and, where a file is specific to a particular set of rubrics, following the orthography of the corresponding typical edition.

fiapps:
I don't know what typical editions you have in mind. A typical edition is a model edition published by the Holy See. Not every reform of the rubrics was accompanied by a typical edition, and simply choosing a breviary to copy doesn't make it a typical edition. Wikipedia tells me that there were typical editions of the missal in 1570, 1604, 1634, 1884, 1920, and 1961. I don't know what the situation is for the Roman Breviary, but the 1568 and 1961 editions are the only ones of which it's possible to buy a photographic reproduction from Libreria Vaticana Editrice. I can guarantee you that you don't want to follow the orthography of the 1568 edition, which still uses medieval abbreviations.

Also, the idea of using different editions for different texts is problematic because of the increasing use of substitutions. A text for older rubrics can well wind up being modified for use with newer rubrics, or vice versa.

It seems to me that the 1961 typical edition is the best model to follow going forward, because it's the most widely used edition, the one clergy can use to fulfill their obligation, and anyone who wants to contribute to the project can obtain a copy. I've been using it as the model for all my changes for a few years now.

I don't propose we make an effort to change the orthography of existing texts, but as corrections are made, let us make them with reference to the 1961 edition.

PR #641, opened 2016-11-26
Code: improve inclusion substitutions function

mbab:
* matching groups can be used
* allow to choose line(s) numbered from 1!
eg. `@:Ant Vespera:1`                                                                        
or  `@:Ant Matutinum:9-10`

+AMDG+

### DIFF
--- a/web/cgi-bin/horas/dialogcommon.pl
+++ b/web/cgi-bin/horas/dialogcommon.pl
@@ -591,9 +591,9 @@ sub do_inclusion_substitutions(\$$)
   # substitute text or select line(s) (numbered from 1!)
   while (($substitutions =~ m{(?:s/([^/]*)/([^/]*)/([gism]*))|(?:(\d+)(-\d+)?)}g)) {
     if ($4) { 
-      local($s) = $4-1;
-      local($l) = $5 ? -$5-$s+1 : 1;
-      local(@t) = split(/\n/, $$text);
+      my($s) = $4 - 1;
+      my($l) = $5 ? -$5 - $s + 1 : 1;
+      my(@t) = split(/\n/, $$text);
       $$text = join("\n", splice(@t, $s, $l));
     }
     else {
--- a/web/cgi-bin/horas/dialogcommon.pl
+++ b/web/cgi-bin/horas/dialogcommon.pl
@@ -588,7 +588,18 @@ sub process_conditional_lines
 sub do_inclusion_substitutions(\$$)
 {
   my ($text, $substitutions) = @_;
-  eval "\$\$text =~ s'$1'$2'$3" while ($substitutions =~ m{s/([^/]*)/([^/]*)/([gism]*)}g);
+  # substitute text or select line(s) (numbered from 1!)
+  while (($substitutions =~ m{(?:s/([^/]*)/([^/]*)/([gism]*))|(?:(\d+)(-\d+)?)}g)) {
+    if ($4) { 
+      local($s) = $4-1;
+      local($l) = $5 ? -$5-$s+1 : 1;
+      local(@t) = split(/\n/, $$text);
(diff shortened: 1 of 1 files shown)


peterdiekertz:
I have no idea whether the code works, but it certainly sounds good!

web/cgi-bin/horas/dialogcommon.pl
```diff
@@ -588,7 +588,18 @@ sub process_conditional_lines
 sub do_inclusion_substitutions(\$$)
 {
   my ($text, $substitutions) = @_;
-  eval "\$\$text =~ s'$1'$2'$3" while ($substitutions =~ m{s/([^/]*)/([^/]*)/([gism]*)}g);
+  # substitute text or select line(s) (numbered from 1!)
+  while (($substitutions =~ m{(?:s/([^/]*)/([^/]*)/([gism]*))|(?:(\d+)(-\d+)?)}g)) {
+    if ($4) { 
+      local($s) = $4-1;
+      local($l) = $5 ? -$5-$s+1 : 1;
+      local(@t) = split(/\n/, $$text);
```
mbab: Corrected.

web/cgi-bin/horas/dialogcommon.pl
```diff
@@ -588,7 +588,18 @@ sub process_conditional_lines
...
+      $$text = join("\n", splice(@t, $s, $l));
+    }
+    else {
+      eval "\$\$text =~ s/$1/$2/$3";
```
mbab: Return to single quotes prevents using matching groups, the feature which can be useful. What is wrong with this?

mbab:
`@:Lectio7:/{:marker1:}/../{:marker2:}`
is equal to
`@:Lectio7:s/.*{marker1:}//s s/{:marker2:}.*//s`

So I'm not sure if it is necessary.

Using line numbers can be very usefull in choose Antifonas or Responsory verses (which are line wise).

web/cgi-bin/horas/dialogcommon.pl
gregordick: `eval` is dangerous and can have unexpected results.  For example, if ```$2 eq '$ENV{PATH}'```, the path will be substituted.  Although this can only be exploited by malicious changes to the datafiles, which are trusted, it's easy to get things wrong by mistake.  If using matching groups is necessary, the regex on line 592 should be tightened up.  Thinking this through, I see it's not even sufficient as it stands.  I'll fix it up.  Since I'll be making that change anyway, I'll take this hunk as it stands.

gregordick:
> So I'm not sure if it is necessary.

You're right.  No new code is needed.

> Using line numbers can be very usefull in choose Antifonas or Responsory verses (which are line wise).

Lines can be joined with tildes, and we should be able to change, for example:

    R. Aspíciens a longe, ecce video Dei poténtiam veniéntem, et nébulam totam terram tegéntem. * Ite obviam ei, et dícite: * Núntia nobis, si tu es ipse, * Qui regnatúrus es in pópulo Israël.

to

    R. Aspíciens a longe, ecce video Dei poténtiam veniéntem, et nébulam totam terram tegéntem. *~
    Ite obviam ei, et dícite: *~
    Núntia nobis, si tu es ipse, *~
    Qui regnatúrus es in pópulo Israël.

and be sure that we won't break anything by doing so.

mbab:
'Tilde' ended lines should be regard as other. It can be even useful.
If someone changes data linked in other files even plain text substitution can break.
So can we keep this intact?

gregordick:
You've persuaded me.  Thanks for your patience!  I would ask that you prefer other substitution methods to line-references where possible, as I still worry that they're fragile, but if they seem to be the most appropriate solution in a given situation, by all means do use them.

PR #648, opened 2016-12-14
Adding November Polish Masses

mmolenda:


DLakomy:
I suggest translating the texts in `[Rank]` (and similar) section, as it's done in breviary. It can be useful one day, eg. if the feasts' titles will be rendered in vernacular on the website.

Similarly, I suggest translating commmorations, I mean the comments starting with `!` sign, for instance:
`!Pro S. Chrysogono Mart.`
In breviary commemorations have been translated in this way:
`!Wspominamy Św. Chryzogona Męczennika`
or
`!Wspomnienie Św. Chryzogona Męczennika`
(we should unify it one day...)

Do you want to include these changes in this PR, or should I merge it as-is and you will do it at another time?

mmolenda:
Good point. Please merge and I'll do that in separate PR.

In Christo
Marcin

2016-12-14 15:10 GMT+01:00 DLakomy <notifications@github.com>:

PR #649, opened 2016-12-14
Polish files: C12 conform Latin changes ++

mbab:
+AMDG+


DLakomy:
Antiphona ad Benedictus is in English during Advent.

PR #654, opened 2016-12-16
Polish Mass December #2

mmolenda:


gregordick:
Many of the sections in these files are identical to those in the Latin versions.  Could you please remove them, so that they will be pulled in automatically from the Latin, and so avoid duplication?

mmolenda:
I removed `[Commemoratio.*]` sections from `Sancti/12-31r.txt`, but I kept all the sections in `Sancti/12-30.txt` as otherwise English version is being pulled.

gregordick:
The fallback to English for missing sections of vernacular translations is by design (falling back to Latin is redundant since it's already in the left-hand column).  But I'll merge this anyway.  Thanks!

PR #659, opened 2016-12-21
Correct various Latin typos, restore English chapter at terce on 12-27

fiapps:
I suggest merging this before 12-27. Otherwise, the English translation of the chapter at terce will be missing.

The change in Sancti/10-10, not so obvious from the diff, is to correct the accent Mortúa to Mórtua.


gregordick:
Thank you, Father.  Sorry that I didn't get to this by yesterday.

PR #666, opened 2016-12-29
Polish files: January

mbab:
+AMDG+


DLakomy:
1. There is an obsolete reference at the end of this line: „aby i onym poczytano było ku sprawiedliwości, **[21]**”.
2. Versum 1 remains in English on January the 1st. I will think why, as I remember we've had a similar problem recently.

mbab:
Ad 1. https://github.com/DivinumOfficium/divinum-officium/pull/666/commits/723ea8b866e0b3dd32371dce085e08d1175ace32
Ad 2. Already in #665

DLakomy:
Thank you. In this case I think that this PR should be merged as soon as #665 is.

PR #672, opened 2017-01-10
Latin/Sancti files and English to conform

mbab:
+AMDG+


gregordick:
Merged.  Sorry for the delay.

fiapps:
One thing to be aware of is that this process of conforming the English to the Latin sometimes eliminates unique translations, i.e., cases where the Latin is the same but the English translation differed. One case of this is St. Hilary (01-14), whose oratio used to have a proper translation, thus:

> O God Who gave to thy people blessed Hilary as a minister of salvation grant we beseech thee that we who cherished him on earth as a teacher of life may be found worthy to have him as an intercessor in heaven…

After 4accbfa, the translation comes from the common, and reads thus†:

> O God who didst give blessed Hilary to be thy people’s minister in eternal salvation grant we pray that that we who have him for teacher of life here on earth may also deserve now that he is in heaven to have him for an advocate…

One can see this as either a good thing (translations are more consistent) or a bad thing (a translation has been changed without consideration of its merits). In either case, I think those interested in the English translation (like @APMarcello3) should be aware that's it's happening. I don't have a strong opinion one way or the other.

† – That's actually how it *will* read once you get my commit that adds the 's currently missing from "to be thy *people* minister".

PR #678, opened 2017-01-26
Polish Mass January

mmolenda:


DLakomy:
Many thanks. Everything seems to be ok. Remember to translate the names one day.

mmolenda:
Thanks. Once the main work is done, I'll do a clean-up PR and will fix all names then.

PR #679, opened 2017-02-01
Polish files: corrections and new Tempora files

mbab:
+AMDG+


DLakomy:
**Epi2-2.txt** and **Epi2-6.txt**:
`Lectio3` is in two versions (Latin and Polish).

PR #685, opened 2017-02-09
Polish: translations of Rank, and Canticum

mbab:
+AMDG+


DLakomy:
`Druga Niedziela Męki Pańskiej czyli Palmowa`

I think a comma is missing before „czyli”. It should be:

`Druga Niedziela Męki Pańskiej, czyli Palmowa.`

DLakomy:
Many thanks!

I just realised that I could merge the PR and just add this comma... I'm sorry for generating the delay.

PR #704, opened 2017-03-10
Polish Mass - adding April feria days

mmolenda:


DLakomy:
Polish `C4bp.txt` - some text are in Latin. Please remove them or translate, so that they be rendered in English.

PR #706, opened 2017-03-16
Fix #413 and other

### LINKED ISSUE #413: Incorrect Benedictio
Benedictio in Matutinum pre1960 is taken incorrectly in polish and probably other languages if Lectio is from Evangelium. 
The reason is hard coded regular expresion in Latin/English eq.:

``` perl
if ($w =~ /\!(Matt|Mark|Marc|Luke|Luc|Joannes|John) [0-9]+\:[0-9]/i) {$a[2] = $benedictio{Evangelica};}                                                  
elsif ($a[2] =~ /(evang|Gospel)/i) { $a[2] = $a[5]; }
```

I can try to prepare patch but it will need add to `Translate.txt` section contains uniform which can be used in regexp. This should match translation form of word Evangelica used in Benedictio and 'Lectio sancti Evangelii secundum'.

In latin

```
[Evangelica_regexp]
Evang
```

In polish

```
[Evangelica_regexp]
Ewangeli
```

In english

```
[Evangelica_regexp]
Gospel
```

Also in german `Translate.txt` file `[Nocturn 3]` 6th Benedictio is duplication of 3rd not proper translation, it is possible solution for a case of above bug but not proper one.

mbab:
+AMDG+

### DIFF
--- a/web/cgi-bin/horas/specmatins.pl
+++ b/web/cgi-bin/horas/specmatins.pl
@@ -890,8 +890,8 @@ sub lectio : ScriptFunc {
   else {$w =~ s/\((Allel[uú][ij]a.*?)\)/$1/isg;}
   if ($dayname[0] =~ /Quad/i) {$w =~ s/[(]*allel[uú][ij]a[\.\,]*[)]*//ig;} 
 
-  #handle parentheses in English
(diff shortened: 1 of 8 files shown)


gregordick:
Great, thanks!

PR #727, opened 2017-04-25
Polish files update

mbab:
+AMDG+


DLakomy:
The PR changes the way the letter 'N' in the lithany is rendered. It's much better now, but only in Polish version. I think it should be changed in other places as well.

PR #735, opened 2017-06-01
Numerous typos, plus orthography corrections

fiapps:
The process of adding accents to prayers (*orationes*) is introducing numerous typos, and is apparently based on an old form of the office, before St. Anthony was a Doctor of the Church. These commits correct all the errors so introduced that I found, but I may have missed others, particularly on feasts not celebrated in 2012, which my test cases cover.

One of the problems introduced by these prayers is changing cael* to coel*. The former is more common in the database, and is the orthography used by the 1962 *editio typica.* There were so many of these that I decided the simplest thing was to switch to cael* everywhere (see commit 651efe0 for more). Other commits below universalize the orthography "solemn*", and use "natalicia" and "intelleg*" more broadly. All of this, plus the corrections above, corresponds to the *editio typica.*


gregordick:
Merged.  Thank you for your patience, Father, and best wishes for today's feast.

PR #770, opened 2017-09-02
Create C1.txt

AbDB1:
début d'une traduction...
In fact, I'd like you to check what's wrong, cause the translations of the psalms in French are done, but when I select Francais in the menu, I get the psalms in English.
If there is something I can do to fix this, just let me know.
In Christo.
And good night.


mbab:
You should change directory name from ```Psalms1``` to ```psalms1```.

PR #835, opened 2017-11-30
Cleanup of Polish translations of psalms (é + BOM)

efati:
* replaced 'é' characters with just 'e'
* removed Binary Order Mark (BOM) from some files

### DIFF
--- a/web/www/horas/Polski/psalms1/Psalm10.txt
+++ b/web/www/horas/Polski/psalms1/Psalm10.txt
@@ -1,4 +1,4 @@
-10:1 W Panu ja ufam: jakóż mówicie duszy mojéj: * Uciekaj na górę jako wróbel?
+10:1 W Panu ja ufam: jakóż mówicie duszy mojej: * Uciekaj na górę jako wróbel?
 10:2 Bo oto grzesznicy naciągnęli łuk, nagotowali strzały swe w sajdaku, * aby postrzelali w ciemności prawych sercem.
 10:3 Albowiem, coś był sprawił, skazili: * sprawiedliwy lepak co uczynił?
 10:4 Pan w kościele swym świętym, * Pan, w niebie stolica jego.
--- a/web/www/horas/Polski/psalms1/Psalm100.txt
+++ b/web/www/horas/Polski/psalms1/Psalm100.txt
@@ -1,7 +1,7 @@
 100:1 Miłosierdzie i sąd * będę śpiewał tobie, Panie:
-100:2 Będę śpiewał, a będę rozumiał na drodze niepokalanéj, * kiedy przyjdziesz do mnie.
+100:2 Będę śpiewał, a będę rozumiał na drodze niepokalanej, * kiedy przyjdziesz do mnie.
 100:3 Chodziłem w niewinności serca mego, * w pośrodku domu mego.
-100:4 Nie kładłem przed oczy swe niesprawiedliwéj rzeczy: * czyniące przewrotności miałem w nienawiści.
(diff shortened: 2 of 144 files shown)


DLakomy:
Thank you for your interest in improving the Polish translation.

I'm unsure whether we should get rid of the „é” letter. There is a significant difference in pronunciation between „é” and „e”. I guess it won't invalidate any rhymes (I've checked quickly, can be wrong), though.

@mmolenda , @mbab May I ask for your opinion?

efati:
We do have "ę" but not "é" in Polish. It does not also represent an accent. I guess it is a mistake.

DLakomy:
It's not a mistake. It's the original typography tha x. Wujek used.

Polish „é” sounds a bit like „i”. I'll try to find an example in poetry where the difference can be spotted.

DLakomy:
> Biegł przebić; wtem Ksiądz dostał z rękawa krócicę:
> "Pal, Tadeuszku! - krzyknął - pal jak w jasną świécę!"
(A. Mickiewicz, „Pan Tadeusz”)
This is wrote with „é” in my printed book (not a very old one ;)). It is common to read „świécę” as „świecę” but it no longer rhymes correctly with „krócicę”.

efati:
OK. Are there any plans to use a modernized version of x.Wujek's translation?
E.g. http://www.madel.jezuici.pl/biblia/Biblia.html – the „C” type of transcription?
I have been working with that translation for some time and I have a set of source txt files.

DLakomy:
There are no such plans. I contacted with Krzysztof Mądel (or someone else whose email I found there)  when I was starting my work here. He advised me to use the „B” version. There were two reasons: it was difficult to get an agreement to use these texts (they are not in Public Domain, it can be difficult to reach the persons responsible for them now). Second, they are no longer Wujek-like, if you know what I mean.

I decided to use texts from Public Domain. It was a safe choice. I didn't ask Pallotins for „Tysiąclatka” as it's controversial among some traditionalists. While Wujek is old in style, it remains a safe choice.

I agree, though, that it would be a good option to have a choice between several different translations. Either here or on another, dedicated project.

efati:
Personally I use the Polish version as a helper text for reading/learning the Latin version. Therefore x.Wujek's translation is necessary for me due to the word-for-word translation style. "Tysiąclatka" is useless in this context. This applies also to the "Edycja św. Pawła" version.
You are right: "B" version is safe. But I personally prefer the "C" version :-)

DLakomy:
`Personally I use the Polish version as a helper text for reading/learning the Latin version. Therefore x.Wujek's translation is necessary for me due to the word-for-word translation style.`

That is exactly my way of thinking ;) I'll check the BOM-thing later. All files should be the same in this matter. I guess all should have BOM but am not sure now. As „é” is correct, I'd opt for leaving it as is. There are several typographical issues we have, for instance: how to write the BMV's name: „Maria”, „Maryja”, „Marya”, „Maryji/Maryi”? There are incosistenscies in capitalisation as well. It should be the same in all places, but it is difficult to satisfy everyone, so it remains as is.

DLakomy:
@efati 
1. About BOM: I'm unable to give an example at the moment, but I recall why I've added this. The heuristical algorithms used to determine which encoding is used have failed to recognise some Polish files.  It was due to „Ś” letter or something like that. I was advised by @gregordick to add BOM, which fixed that.
2. About „é” I think we can agree it's not a typo. I'm in doubt whether we should change it, in this case. Would you mind my rejecting your Pull Request? I think it would be better to stay close to the source text, wchich Gregor asked me to do.

efati:
OK, reject it.

PR #841, opened 2017-12-06
Missing Ps 94

efati:
- Added missing Ps 94.
  The contents comes from Latin/Psalterium/Invitatorium1.txt supplemented with missing verses 6 and 7.
- Ps 4: missing newline at the end of file


DLakomy:
A little explanation (efati explained it to me some time ago via email):
Ps. 94 is added there for consistency. It will be easier to handle any import of texts from our project.

efati:
Exactly!

PR #843, opened 2017-12-06
Cleanup of French translations of psalms

efati:
- Fixed numbering labels in Ps 46 and 47
- Removed remaining spaces from ends of lines

Remaining problem: Ps 1 has 6 verses while the Latin version (https://github.com/DivinumOfficium/divinum-officium/blob/master/web/www/horas/Latin/psalms1/Psalm1.txt) has 7. Can someone fix this?

// I guess you meant Latin has 7 - DL


efati:
Yes, Latin version of Psalm 1 has 7 verses.

PR #884, opened 2017-12-14
Numbering of verses in Psalms 78, 79+refs, 80 and 82

efati:
Additionally the German version of Ps 80 (which is Latin in fact) was converted from CP1252 to UTF-8.


DLakomy:
@peterdiekertz I think that German files which are Latin in fact should be removed, so that they be rendered in English.

PR #903, opened 2017-12-20
Verse numbering errors and typos

fiapps:
I caught some errors in the new verse numbering that were causing verses to appear in the wrong psalm section. There are also a few typos, and I've eliminated a duplicate section from the Latin `Prayers.txt`. The duplicate section had no effect, but it's tidier to eliminate it.


efati:
Concerning verse numbering: you are right. I have not applied the update of Latin version to other translations in case of Ps 144.

PR #958, opened 2018-04-10
Minor fixes in Polish Mass files

mmolenda:


web/www/missa/Latin/Sancti/12-25m3.txt
```diff
@@ -74,4 +74,4 @@ nobis generatiónis est auctor; ita et immortalitátis sit ipse largítor:
 $Qui tecum
 
 [Ultima Evangelium]
-@Tempora/Epi0-0:Evangelium
```
mmolenda: There no such file - `Tempora/Epi0-0.txt`

PR #966, opened 2018-05-04
Corrections mostly to Latin

fiapps:
* Most of these commits correct typos.
* The Annunciation had lost the versicle at vespers. I've restored it with a reference.
* The new oratio for St. Joseph (1962 rubrics) had the wrong conclusion. I've corrected it for all languages.

As a rule, I check my corrections with the *editio typica.*


rl1234567890:
horas/Latin/Sancti/03-25: missing versum 3
Do we need to add it to the translations?

fiapps:
No. The reference I added to the Latin will be picked up by the translations as well, and resolved in the language of the translation.

PR #993, opened 2018-07-28
Revert "code: make Ordinarium Prima, Tertia and Completorium paschal files obsolete"

APMarcello3:
Reverts DivinumOfficium/divinum-officium#992


mbab:
May I know the reason?

PR #1000, opened 2018-08-09
code: make Ordinarium Prima, Tertia and Completorium paschal files obsolete, fixed for perl 5.10.1

mbab:
+AMDG+


fiapps:
Is this compatible with Perl 5.8, or does it require 5.10?

mbab:
@fiapps: I used construction like `%{x}{y}` which breaks in perl 5.10.1 (version used at divinumofficium.com), I don't suspect any further problems with earlier version.

Issue #938, opened 2018-01-29
New Polish translations


efati:
Polish translations of Psalms and other canticles come from the public domain translation done by Fr. Jakub Wujek (https://pl.wikisource.org/wiki/Biblia_Wujka_(1923)). It is a very respectable and important translation which is moreover very useful because it strictly follows the original Vulgate Latin source. However, the transcription (called „B”) is quite old and uses language forms which becomes more and more unintelligible. I have found another translation or another version of the translation done by Fr. Stanisław Styś (https://pl.wikipedia.org/wiki/Stanis%C5%82aw_Sty%C5%9B). It is a revised (not to say modernized) version of the Fr. Wujek translation. It is still very close to Latin original, but can be read as easily as any contemporary text.

I have recently contacted the provincial of the Jesuit order who have the copyrights for the translation and I'm happy to announce that they gave me permission to use the translation in Divinum Officium project. Therefore I would like to add the translation of Fr. Styś as an alternative to the existing one. It raises a couple of questions:

1. Following the recent addition of Pius XII psalter should I use for example `web/www/horas/Polski/Stys` directory for psalms?
2. An alternative solution would be to create a new top-level language, e.g. _Polski2_ or _PolskiS_. It may be necessary due to the differences of language style in other texts like antiphones and hymns. The new language _Polski2_ could default to _Polski_ in case of missing source files.
3. Another problem comes from different styles of writing personal pronouns. In Polish we used to write them with a small letter in the past, and nowadays we start with a capital letter. Personally, I prefer the new notation because we are used to it. Moreover, it is simple to programmatically (and dynamically) change all „On/Twój/Jego/…” into „on/twój/jego/…”. The inverse is impossible to automate.

What is your opinion?

DLakomy:
Ad. 1. and 2. I haven't got an opinion; nonetheless, I would like to avoid showing two different „Polski” entries in the menu without any nesting/subdirectory. It would look strange to me.
Ad. 3. As I understand you would like to capitalise personal pronouns addressing the God (and the BMV, right?) in the source files and add an option to lowercase them? If so, I like the idea.

efati:
Ad. 2. I think of a whole new directory structure `web/www/horas/Polski2` with subdirectories `psalms1`, `Psalterium` etc.
Ad. 3. Precisely.

efati:
There is also one more solution:

[...]

PR #1030, opened 2018-11-17
Multi fallback implementation (#938) - discussion needed

DLakomy:
I've implemented a fallback mechanism for #938 according to @fiapps suggestion ([link](https://github.com/DivinumOfficium/divinum-officium/issues/938#issuecomment-434046885)).

There is a bug, though. In some places the parameter `$lang` has `_` replaced with ` ` (a space), notably `mpopup.pl`. Owing to this fact, if I click the link to the `Communio` popup (while reading `Sancta Missa`), using the `Polski_New` language (without the proper translation of the communio among the files), I get English instead of Polish.

I see two ways of fixing this:
1. Change the separator to `-` (easy, but not elegant).
2. Remember to replace a space with and underscore where needed in `$lang` (at the moment I know only about `mpopup.pl`).
3. (best) Find a reason why `_` is encoded as space and fix it.

### DIFF
--- a/web/cgi-bin/horas/dialogcommon.pl
+++ b/web/cgi-bin/horas/dialogcommon.pl
@@ -304,6 +304,13 @@ sub setupstring($$$%)
       # English layers on top of Latin.
       $base_sections = setupstring($basedir, 'Latin', $fname, 'resolve@' => RESOLVE_NONE);
     }
+    elsif ($lang =~ /_/)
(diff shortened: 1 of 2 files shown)


APMarcello3:
> 1. Change the separator to `-` (easy, but not elegant).

Let's do this solution for now and I'll merge the pull. Thank you!

PR #1045, opened 2018-12-13
New translations of Polish psalms

efati:
This PR is a result of #938. I assume that this is an initial part of long term effort of preparing more contemporary Polish translations of texts of the breviary that will preserve Tradition (Fr. Wujek) and be more accessible to contemporary readers.

These translation is an update of Fr. Jakub Wujek translation done by Fr. Stanisław Styś (https://pl.wikipedia.org/wiki/Stanis%C5%82aw_Sty%C5%9B).
The verses were rearranged to fit the layout of breviaries.
This package contains all 150 psalms + Quicumque.

### DIFF
--- a/web/www/horas/Linguae.txt
+++ b/web/www/horas/Linguae.txt
@@ -5,3 +5,4 @@ Francais
 Italiano
 Magyar
 Polski
+Polski-Newer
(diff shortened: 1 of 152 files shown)


DLakomy:
Many thanks. It seems to be good. I'll try to test it locally later today.

However, I think one technical thing is missing. In the file `divinum-officium/web/www/horas/Linguae.txt` the line `Polski-Newer` must be added, so the new language be selectable.

efati:
I have noticed that but postponed the update for the software maintainers. Should I include it in this update?

DLakomy:
I think it can be included in this update. I don't think we should wait for more translations before making the new psalms publicly available. Of course I don't force my opinion.

efati:
OK. Done.

PR #1064, opened 2019-01-16
Typo and orthographic correction

fiapps:


APMarcello3:
The error on English/Tempora/Adv1-0 has been rectified and the orthography of "obediens" will be used in correspondence with the texts found in the editions of Maison Mame and Desclee.

PR #1070, opened 2019-02-08
Latin/* - replace English words in the titles of the feasts with Latin

mbab:
+AMDG+


fiapps:
The name of the PR is somewhat misleading: it corrects the `[Rank]` section of some files, but does not actually change the rank of the feasts. Instead it replaces English words in the titles of the feasts with Latin. I approve of the PR, but just wanted to clarify what it contains.

PR #1117, opened 2019-04-13
Add Portugues Folder & the psalms

ofrades:


fiapps:
English is automatically substituted when there is no translation in the language requested. There is no need to add files containing English text to a Portoghese folder.

PR #1120, opened 2019-04-16
Dockerfile for running app in Docker [WIP]

benyanke:
Working on a Dockerfile for the app to allow both easier development (for those who wish to use it) and future deployment.

Would love input if others have docker experience or thoughts.

Left todo before merge:

 - [x] working Dockerfile
 - [ ] cleanup `docker/apache/httpd.conf`
 - [ ] Add init system
 - [ ] Look at env vars / config needed to ensure it's fully covered
 - [ ] Add healthcheck
 - [ ] Add documentation to `README.md`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,38 @@
+FROM        httpd:2.4-alpine
+# FROM        perl:5.28
+MAINTAINER  Ben Yanke <ben@benyanke.com>
+
+###################
+# Divinum Officium
+###################
+
+# Creates a fully running application server for Divinum 
+# Officium, also enabling easy development without configuration
+# or installing dependencies
+
+# Last modified 4/12/2019 by Ben Yanke <ben@benyanke.com> 
+
+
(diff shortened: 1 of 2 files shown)


DLakomy:
I've created a Dockerfile some time ago, with similar purposes. What do you think about it?

The address: https://hub.docker.com/r/dlakomy/perl-fcgi/

benyanke:
@DLakomy I've used nginx in the past before, and I have found it to sometimes be ligherweight when serving a ton of static files - I haven't done much benchmarking with perl before.

hason:
Although I prefer nginx in other projects, it is better to use Apacha in this case because one process to run the project is enough. It will not be necessary to solve the init system.

benyanke:
In the case of the init system, it's not because of a massive number of processes, or services required. It's to improve things like signal handling (something most programs aren't designed to do as an init process) and zombie reaping.

I have a patch just about ready adding `dumb-init` - it's pretty seamless, and handles these things without getting in the way, just by specifying it as an entrypoint before the cmd.

PR #1122, opened 2019-04-17
Add docker-compose for local development and simplify docker configuration

hason:
I have created multi stage Dockerfile for development with docker-compose without files from repository (48M) and production with all necessary files (168M instead of 245M of the the current version of Dockerfile).

ping @benyanke @DLakomy


benyanke:
Thanks for the additional info - I did want to add a docker-compose file at some point.

Out of curiosity, is there a specific goal you're accomplishing by using `minidocks/perl`? I prefer to try and not use non-core base images when possible, to keep things as up to date as possible. I certainly can adjust the existing one to only grab the web files - that is a good suggestion. I'll be merging these two ideas soon.

Additionally, instead of using the from-lines, I'd rather break out into two dockerfiles if we're going to have separate dev prod, to follow best practices.

hason:
@benyanke I'm a maintainer of `minidocks/perl` image. This image is the smallest with perl and apache2. There is no core docker image with perl and apache2. Another advantage of `minidocks/perl` is that the main process is controlled by tini https://github.com/krallin/tini.

benyanke:
Thanks for the contributon, hanson. At this point, I think we'd like to remain on the core docker library images to reduce dependency sprawl, but I do appreciate the work! I'll be merging in some of your changes soon wit what we previously had and PRing that in the next day or two

PR #1126, opened 2019-04-20
Improve dockerfile and docs

benyanke:
resolves #1121, adding documentation and cleaning up the dockerfile to refactor on the perl base image.

PR #1128, opened 2019-04-20
Use slim perl docker image

hason:
Fix #1126: 340 MB instead of 1.02 GB.


benyanke:
Thanks! This should slim the image down some (no pun intended)

Issue #1658, opened 2020-05-05
Missing form labels and alternative text


martineizayaga:
- /cgi-bin/horas/Ckalendar.pl
    - [x] add 2 form labels
- /cgi-bin/horas/Cofficium.pl
    - [x] add 5 alternative text
    - [x] add 6 form labels
- /cgi-bin/horas/kalendar.pl 
    - [x] add 1 form label
- /cgi-bin/horas/officium.pl
    - [x] add 5 alternative text
    - [x] add 6 form labels
- /cgi-bin/missa/Ckalendar.pl
    - [x] add 2 form labels
- /cgi-bin/missa/Cmissa.pl
    - [x] add 7 form labels
- /cgi-bin/missa/kalendar.pl
    - [x] add 1 form label
- /cgi-bin/missa/missa.pl
    - [x] add 7 form labels

martineizayaga:
Could I be assigned this issue?

APMarcello3:
Please proceed.

APMarcello3:
This commit caused a serious crash.

martineizayaga:
Why was the pull request reverted? If you can, please revert the revert. I’m done with this issue

APMarcello3:
It would be very helpful to us if you could provide a more detailed explanation of what your changes are and what they are aiming to accomplish. I am not sure I am seeing what the aim is of the "form labels".

martineizayaga:
Ah! For sure. I was following the accessibility web standard that every form input should have a corresponding label.
https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-explicitly

I then hid the labels visually, that way the design is the same, but people with screen readers can now tell what kind of input they are interacting with. I hid the labels with the class `.offscreen` that's taken from this url: https://www.30secondsofcode.org/css/s/offscreen/

martineizayaga:
I also made sure to add empty alt text's for decorative images.
https://www.w3.org/WAI/tutorials/images/decorative/#image-used-as-part-of-page-design

APMarcello3:
Okay. If this is ready to implement, could you submit a new pull? Thank you.

PR #1659, opened 2020-05-05
Missing form labels and alternative text #1658

martineizayaga:
Missing form labels and alternative text #1658


APMarcello3:
This commit has caused the site to crash.

PR #1699, opened 2020-05-14
Typo

dokeefe9999:


dokeefe9999:
Exulted shows up as variation and is probably archaic form of exalted.  1962 English missal has exalted.

PR #2329, opened 2021-01-27
Fix bible refs

mcsitter:
Bible refs have all kinds of formats. Unifying them should make parsing easier and could show places in which wrong bible verses are used. Also sometimes thing are omitted, for example the book and verses are present, but the chapter is missing. There is more to come!


APMarcello3:
The period at the end of the reference should also be eliminated.

mcsitter:
I made these changes while developing a regex, which would help with this, but some things needed to be changed manually to narrow down the possibilites. I want to update all bible refs under Lectio and Evangelium section in one go. Refs in other section may come later. I would also like to make sure that the ref is the same across all languages for the same reading. Also standardize the different ways to write John, Joh, Jo, Juan, Mt, Mat, Matt, Matth and more! Might do the name standardising in a different commit.

Issue #626, opened 2016-11-02
Separate Logic from Display


edbrannin:
Would you be open to PRs that start to separate logic from display?

For example, instead of interleaved "Determine today's reading" and "print HTML for the reading", the result of this would be one function that determines what content should appear on the page, and another function that injects that content into an HTML file, probably with [Template::Semantic](http://search.cpan.org/~tomita/Template-Semantic/lib/Template/Semantic.pm) or something similar.

(Long-term, I'd like to try making a more modern-styled version of this site, and slowly doing this would be the first step to having a JSON API.)

gregordick:
Yes, please!  The current implementation is extremely fragile, so I would suggest making very small, incremental changes.  You might also be interested in the work on the `calcalc` branch that I'm determined to merge one day.

For regression testing, you might be interested in the scripts in `regress` directory.

osoftware:
Is that branch actively developed? The last pushed commit is one year old.

gregordick:
I haven't had time to work on it recently, but it hasn't been abandoned.  The present difficulty is that it requires a change to the data-file format.  I started doing it manually, but I can't keep up with the changes on `master`; when I get time, the next thing I'll do is to automate the change to the files so that I can do it atomically.  Pull requests are always welcome, but if there are people interested in working on that branch, it might be helpful if I were to write some documentation first.

edbrannin:
Do you think it would be a better idea to do a change like this against `master` or `calcalc`?

gregordick:
`master` would be better.  `calcalc` doesn't really touch the presentation logic (such as it is!), so I would expect minimal merge pain.

osoftware:
I would gladly help with creating a converter to the new data format.

gregordick:
Thank you: that would be very welcome.  If you're still interested (sorry for my delay...) I can document the required transformation.  The essence of the change is that separate offices on a single day need to live in separate files.

osoftware:
Still interested.

osoftware:
bump

DLakomy:
@osoftware I tried to contact with @gregordick, with no success. I can suggest reading this wiki page: [link](https://github.com/DivinumOfficium/divinum-officium/wiki/CalendarCalculation). I don't know if it's still valid, unfortunately. I'm sorry I can't help more. I don't want you to remain without any answer.

ewcole:
[...]

PR #2436, opened 2021-03-16
[Experiment] Use Template::Semantic to isolate logic

malukenho:
I need a review on it, as well as opinions on that. I know that we have an issue https://github.com/DivinumOfficium/divinum-officium/issues/626 from 2016. I'm not sure if that still applies.


malukenho:
I'm not even sure what is the future plans for this project. Is there any place I can look into it?

gregordick:
Thanks for the PR.  Templating is a good thing, but I think that the presentation, logic and database are so intertwined that retrofitting it is likely to be prohibitively difficult.  Then again, I'd be glad to be proven wrong.  Before we pull in another dependency, I think if we're to pursue a project of this sort, we should try to push the templating deeper and see which bits are difficult.  If you're interested in looking into this, I'd be glad to review it.

As it happens, I've been working on an off for a few years on a [proof-of-concept implementation of a Python package](https://github.com/gregordick/divinum-officium-alterum) for generating offices from the DO datafiles, and I've just started on the templating in the past week or two (but haven't pushed anything yet).  This is very much a side project and might never go anywhere, but I hope that it will provide a useful point of reference for discussions of possible future directions.

PR #2467, opened 2021-04-04
Adding TemporaM files, spanish version

jesav:


mbab (commented):
Thank You very much for contributing to Monastic office.
Please consider to always use current data, eg. Pasc4-2 is just link to non Monastic file now and there is need to translate anything.
On other side there is no need to translate LectioE as it is now taken from 'missa' files (I'm aware there is no missa dir in Espanol but better is to put it there).

+

jesav:
Thanks for your comment. If that is the case, I prefer to complete the spanish version of Monastic office once it has been finished in another reference language (English or Latin). Instead, I will start to find suitable translations for the spanish version of the Mass.

Issue #254, opened 2015-04-06
Make title (and other?) translations more generic


GoogleCodeExporter:
```
For example:

> In the Magyar version the 'Lectio 1' title is 'Olvasmány 1'. It is not 
correct, we
> should use this format: '1. olvasmány'. It would be great if someone from 
the team,
> could fix this problem.

```

Original issue reported on code.google.com by `igregord` on 8 Nov 2014 at 11:05

PR #2574, opened 2021-08-02
code: implement #254

mbab:
+AMDG+


mbab:
> It works!
> 
> Do we have publications in the other languages which can suggest if this new feature ought to be used when printed?
There is no polish translation (just divinumofficium.com :-)
I spotted that old books use roman numbers maybe we can abandon sprintf and use own function with can support roman numbers as %r and %n for arabic.

PR #2590, opened 2021-08-12
Match English to Latin in Communion Antiphon

dokeefe9999:


ajmalton:
Strangely: this commit appears in my github app to consist of two files which both had every line replaced by the same line again.

What _really_ happened?

dokeefe9999:
All I did was replace "him" with "Christ the Lord" in the Communion antiphon.

I have a new laptop. I wonder if I set up Git line feed/return line endings incorrectly?  I believe I set them up for no conversions either way.
________________________________

ajmalton:
That seems very possible. Also I am not convinced your master is properly based on our upstream.

Your file C6bp.txt does not exist in master. Your file C6b.txt contains a whole Commune: but in master it's just a one line reference to C6.

Can you clean up your history and reissue the PR?

dokeefe9999:
Will do. Thanks.
________________________________

ajmalton:
But looking more carefully, perhaps, I see that you made that one change, which indeed make the English clause a correct translation.

However, it does still appear in git that the whole of both files have been changed, with most lines apparently identical before and after. Obviously that's not desirable in general, even if here it might not matter much.

Can you see what's causing this and fix it? (Does the PR look like that to you?)

dokeefe9999:
I'm reinstalling Git on my machine, this time converting Unix line feeds to Windows style and then back again. I'll also check on the existence of these files once I reclone. Maybe I cloned from the wrong source originally.
________________________________

dokeefe9999:
Funny, I just recloned and C6bp.txt is still in there.  When I look at the github site I also see it.

[cid:8706bd0a-4158-4045-9714-69513ad56ab6]

________________________________

PR #2591, opened 2021-08-12
Change English to match Latin in Communion Antiphon (correction of #2…

dokeefe9999:
…590)

This looks better now.  I think the problem was the cntl/linefeeds.

Sorry for the confusion.


ajmalton:
I suppose you are using Windows... perhaps you would find core.autocrlf=true a useful config setting in git.

dokeefe9999:
Does that automatically detect and correct the line endings?
________________________________

ajmalton:
It does, yes.  See [here](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration).

PR #2631, opened 2021-09-27
Cuius->Quorum

KPFC:
S. Eustachii et Sociorum Martyrum is a group of martyrs, so plural in the Benediction.


KPFC:
I just noticed it should also be changed from S.->Ss., otherwise it doesn't make sense.

PR #2919, opened 2022-05-28
code: consolidate officium & Pofficium

mbab:
+AMDG+


mbab:
@fiapps should review this perheps

PR #3008, opened 2022-09-24
Polski: update Invitatorium

mbab:
+AMDG+


DLakomy:
My response to the review request (it's already merged, though): the new text sounds good and is more faithful to the original. Good change.

PR #3191, opened 2023-03-30
transfer tables based on Easter date

mbab:
details in commit message

+AMDG+


web/cgi-bin/DivinumOfficium/Directorium.pm
```diff
@@ -0,0 +1,195 @@
...
+  load_data_data unless defined $_dCACHE{loaded};
+  my $key = shift;
+
+  defined $_dCACHE{$key}
+}
+
+sub load_transfer_file {
+  my $name = shift;
+  my $filter = shift;
+  my $type = shift;
```
gregordick, line 50: I think it would be worth sanitising these parameters, even though we expect them to come from the datafile.  Maybe enforce that each matches `/[a-z0-9]/i`?

web/cgi-bin/DivinumOfficium/Directorium.pm
```diff
@@ -0,0 +1,195 @@
...
+  } else { # whole year
+    @lines;
+  }
+}
+
+sub load_kalendar {
+  my($version) = @_;
+  my $cache_key = "kalendar:$version";
+
+  my($filename) = "$datafolder/Kalendaria/$_data{$version}{kalendar}.txt";
```
gregordick, line 68: Please sanitise as above.

web/cgi-bin/DivinumOfficium/Directorium.pm
```diff
@@ -0,0 +1,195 @@
...
+
+  unless (is_cached($cache_key)) {
+    my $isleap = leapyear($year);
+    my @easter = geteaster($year);
+    my $easter =  $easter[1] * 100 + $easter[0];
+
+    my @lines = load_transfer_file($easter, $isleap, $type);
+    if ($isleap) { # load Jan & Feb from next file
+      $easter++;
+      $easter = 401 if $easter == 332;
```
gregordick, line 106: I see that this version of the patch doesn't have `426.txt`.  Is this intentional?

web/cgi-bin/DivinumOfficium/Directorium.pm
```diff
@@ -0,0 +1,195 @@
...
+
+### get_kalendar($version, $day)
+### get filename for sancti day
+sub get_kalendar {
+  my($version, $day) = @_;
+  my $cache_key = "kalendar:$version";
+
+  load_kalendar($version) unless is_cached $cache_key;
+
+  $_dCACHE{$cache_key}{$day}
```
gregordick, line 132: I'm a bit nervous about returning references straight from the cache, both here and in the other similar functions.  Have you checked that the callers don't modify the objects returned here?  As an experiment, if you make all of the cached values [Readonly](https://metacpan.org/pod/Readonly), recursing down nested references, does everything still work?

mbab: both get table functions will be discarded.

web/cgi-bin/DivinumOfficium/Directorium.pm
mbab: In next PR, there will be checks of $version in callers. I suspect it suffice.

web/cgi-bin/DivinumOfficium/Directorium.pm
mbab: $version will be check in all load_ functions.

PR #3195, opened 2023-04-12
code changes around transfer by easter

mbab:
- code: include transfered in Directorium and other improvments
- code: work around St. Mathias fest & vigil in leap years

+AMDG+


gregordick:
Thanks.  LGTM.

PR #3237, opened 2023-05-23
Remove client-side redirects in favor of server side redirects

benyanke:
The loadbalancer has been adjusted to do server-side redirects from http -> https, so these client-side redirects are now superfluous.


benyanke:
Note - waiting to merge for a few hours, to ensure the new LB has DNS propagated globally.

Issue #3453, opened 2023-12-04
Vernacular commemorations appearing in Latin


APMarcello3:
![Divinum Officium Laudes.png](https://github.com/DivinumOfficium/divinum-officium/assets/11463422/97e2aae4-758e-4522-bbdd-8a8bb0b85639)

From a correspondent. The French translations are given in the respective files but these appear in Latin. I suspect the new commemoration logic may be causing this issue.

@FAJ-Munich would you kindly investigate?

FAJ-Munich:
Just saw that the Latin [Commemoratio] in other cases also supersedes the Vernacular... 
I'll investigate further in the code

APMarcello3:
@benyanke could perhaps give some insight as he is working on this issue.

benyanke:
Hi All - just rolled the deployment manually, and will continue to work on improving the deployment automation.

PR #3455, opened 2023-12-04
Follow up to #3454

FAJ-Munich:
Replacing names in Commemorations in the Vernacular #3453


FAJ-Munich:
This is still not the perfect solution to all the oddities I have just discovered tonight, but should at least address the raised issues discovered by others. More to follow.

In addition, this should also (hopefully completely) solve the issues with the "Compare Office" functions where,  so far, some Antiphones, Hymns etc... got mixed up between two compared offices from time to time. Also, it should be able to fix the headline for the Comparison when vernacular language is selected and the office names are actually translated.

PR #3472, opened 2023-12-09
Fix Latin Dec 29 and add Ukrainian translation of Dec 30 and Sunday within Christmas Octave

theodotk:
- **attention** December 29 in 1960 rubrics was missing Christmas texts. I fixed it by importing them from `12-30o`. Please correct if this is not the best import
- Add Ukrainian Dec 30 translation
- Correct Latin Dec 30 and Sunday within Nativity Octave
- Add back Ukrainian Nativity Masses (silly git mistake on my part)
- Add Ukrainian Sunday within Nativity Octave


FAJ-Munich:
The Office for Dec 30 is Temporal only and has been moved and renamed to the Tempora folder: Nat30.txt and Nat30o.txt respectively. I'm the Sancti folders, 12-30 files are obsolete. Thank you.

theodotk:
@FAJ-Munich is this folder migration happening to some other days? To know what to do if I stumble upon a similar situation again

FAJ-Munich:
@theodotk Hi,
I do not think so. The folder migration was for the days after Christmas only such that (for the office) it has been easier to handle the various rubrics structures on when the Sunday in the Octave of Christmas is to be celebrated and which readings are "winning". Since missa is tied to the same calendar and transfer data files as the office, this has had a spillover effect. 
The only other days, where the split between sanctoral and Tempora could have been necessary too, are January 2–5. However, I have had no trouble getting the desired results without moving around the existing files.

theodotk:
@FAJ-Munich thank you for the answer!

theodotk:
@FAJ-Munich I checked, and for January 3 and 4 the Mass for 1960 and 1955 rubrics can't find propers at the current version of the `main` branch. I'd try to correct it myself in a PR, but there are no files for these days in `Tempora` for neither `horas` nor `missa`, and the files in `Sancti` have all the references, so I'm not sure what to correct there

Issue #3539, opened 2023-12-30
Lections if Nat1-0 falls on December 26-28 (1960 rubrics)


APMarcello3:
Applicable for years C, D, E. Not appearing correctly, for instance, on 12-26-2010. See rubric.

![Screenshot from 2023-12-30 09-50-32](https://github.com/DivinumOfficium/divinum-officium/assets/11463422/50c8fc57-f632-4966-92bb-dd510f7c21fd)

PR #3541, opened 2023-12-30
1960: issue #3539 and some issue related to #3536

FAJ-Munich:
+ Addressing the underlying code issue for Rubrics 1960 for Occuring Scripture when Sunday falls on December 26–28 (correction to e07f9bc65) 

+ after merging #3535 and #3536, Divino Afflatu and 1955 became irresponsive on my local server. Removing the fifth element in `data.txt` resolved the issue for now. There seems to be more work necessary, @d-a-s , to make the `$base` variable in `directorium.pl` work; in the short term, I couldn't figure out what exactly.


d-a-s:
@FAJ-Munich Sorry, I didn't notice that some rows already had a fifth value. Was the fifth element in `data.txt` being used for something else?

The problem would have been that the new `base` code was expecting a valid `version` to be in the fifth value, and there was no other version in the `data.txt` named `DA` or `1955`. I can implement a second check to validate that, so that it will fail more gracefully if that happens again.

FAJ-Munich:
> @FAJ-Munich Sorry, I didn't notice that some rows already had a fifth value. Was the fifth element in `data.txt` being used for something else?

No worries; I have no idea where these 5th values had come from; probably added by accident at some point in the past; and I do not see how these values could have been used before so it shouldn't be an issue for now.

PR #3642, opened 2024-02-23
align portuguese text with latin

ofrades:
@APMarcello3 I have concatenated the latin passion text, hope that's fine.


APMarcello3:
> @APMarcello3 I have concatenated the latin passion text, hope that's fine.

Kindly remove concatenation.

ofrades:
@APMarcello3 I have removed the concatenation

Can you explain to me what is the purpose of the split in just this cases? I ask because I them concatenated  in other places.

APMarcello3:
> @APMarcello3 I have removed the concatenation
> 
> Can you explain to me what is the purpose of the split in just this cases? I ask because I see them concatenated  in other places.

The Palm Sunday Gospel is split as it can be read in an abbreviated form if the priest celebrates >1 Mass that day.

PR #3682, opened 2024-03-17
Fixed Asperges me and Vidi aquam that were in english.

ManyGrek:
The asperges me, and vita aquam were in english instead of spanish. This texts were changed for their proper translation in spanish.


ManyGrek:
This parts were in english, so I changed it to spanish. That is all.

PR #3722, opened 2024-04-22
use &antiphona_finalis in CompletoriumOP

mbab:
+AMDG+

### DIFF
--- a/web/cgi-bin/horas/horas.pl
+++ b/web/cgi-bin/horas/horas.pl
@@ -403,26 +403,29 @@ sub Benedicamus_Domino : ScriptFunc {
 #return the text for the appropriate time
 sub antiphona_finalis : ScriptFunc {
   my $lang = shift;
-  my %ant = %{setupstring($lang, "Psalterium/Mariaant.txt")};
-  my $t = '';
+  my $name;
 
-  if ($dayname[0] =~ /adv/i && $winner{Rank} !~ /In Nativitate Domini/i) {
-    $t = $ant{'Advent'};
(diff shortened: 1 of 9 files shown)


web/cgi-bin/horas/horas.pl
```diff
@@ -403,26 +403,29 @@ sub Benedicamus_Domino : ScriptFunc {
...
+    $name = 'Quadragesimae';
   } elsif ($dayname[0] =~ /Pasc/) {
-    $t = $ant{'Paschalis'};
+    $name = 'Paschalis';
   } else {
-    $t = $ant{'Postpentecost'};
+    $name = 'Postpentecost';
   }
-  $t = '#' . translate('Antiphona finalis BMV', $lang) . "\n$t";
+  my $t = %{setupstring($lang, "Psalterium/Mariaant.txt")}{$name};
```
gallizoltan, line 426: Hi! When I try to run the project, the browser shows an error for this line:
```
Software error:
syntax error at divinum-officium/web/cgi-bin/horas/horas.pl line 338, near "}{"
BEGIN not safe after errors--compilation aborted at divinum-officium/web/cgi-bin/horas/horas.pl line 338.
```
It worked before this modification. Could you please help me, how could I make it run again?

mbab: Does this is error occurs with current master tip?

gallizoltan: I think I managed to figure it out. I have an old setup with an ancient Perl version, which does not support postfix dereferencing. I'm trying to solve it without any side effect: #3893

PR #3740, opened 2024-05-06
Pull-MatthieuFiq

APMarcello3:


MatthieuFiq:
@APMarcello3 
the confiteorM no longer mentions "to our blessed father Saint Benoit" in the French translation.
Before this sentence was not translated (remained in English). I made this modification in Prayer.txt but perhaps not correctly
In link maybe with #3759

PR #3755, opened 2024-05-12
fix regression in PrimaOP.txt

happy-lisper:
Prime is showing the macro name instead of actual prayer. Likely a regression.


happy-lisper:
I am new to the project and was also wondering if MisereaturOP/ConfiteorOP/etc could be removed from web/www/horas/Dansk/Psalterium/Prayers.txt

PR #3765, opened 2024-05-14
Revert "Reduced 1955: cosmetics for "neglected" rubrics"

APMarcello3:
Reverts DivinumOfficium/divinum-officium#3764


APMarcello3:
Causes site to crash.

Issue #3654, opened 2024-02-27
Pre-1955 paschal vigil (DA and earlier Roman) missing Scriptural references


MRoth1910:
The scriptural references for the prophecies and readings, including those of the Mass, are missing.

FAJ-Munich:
Have you tried activating the rubrics and does the issue persist? Due to its special character, the scripts work quite differently for the paschal vigil.

PR #3769, opened 2024-05-15
Revert Quad6-6 unintentional change

ofrades:
This change was added in [Closes issue #3564](https://github.com/DivinumOfficium/divinum-officium/commit/050925e473c21aee9c386459415ae8005c850a5a) but I am not sure if this was intentional because it's not a pattern used in the rest of the repo. Usually, we have `!Tractus` and Verses also have only one `!`


ofrades:
@APMarcello3, you closed the PR without any comment, which leads me to think this change was indeed intentional. Can you explain why this file should be different from the rest?

APMarcello3:
https://github.com/DivinumOfficium/divinum-officium/issues/3654

PR #3780, opened 2024-05-16
Update Prayers.txt

MatthieuFiq:
Adding newlines to match translations and align the text
Capture


mbab:
I can't see _Pater noster_ printed with line breaks in any Breviary.

MatthieuFiq:
In this case, I purpuse to cancel my modification and to delete line breaks in french traduction.

PR #3790, opened 2024-05-18
Versicle "V." → "℣." & Response "R." → "℟"

Geremia:


APMarcello3:
New characters do not appear in red.

fiapps:
I know this PR had been reverted already, but lest there be a second attempt, I also think this is a bad idea because it requires people who contribute translations to have more technical knowledge in order to produce the necessary characters. It's easy to type "V." and "R." with an ordinary keyboard, and since `webdia.pl` lines 440–441 already convert V. and R., there's no benefit to the end user to making source files more complicated to prepare.

Geremia:
@fiapps 
> `webdia.pl` lines 440–441 already convert V. and R.

Not all the V. and R. are converted, though. I see a mixture "V.", "℣.", "R.", and "℟".

fiapps:
@Geremia I would consider that a bug, since the code attempts to convert them. So, just report the cases you find.

PR #3865, opened 2024-06-14
Fixed erroneous verse separation in ps. 21

xmarteo:
See Antiphonale Romanum, editio typica, p. 149:
https://gregobase.selapa.net/source.php?id=9&images=1&index=6

Verse begins with "In te projectus sum ex utero"


APMarcello3:
The original separation of lines is what is indicated in the Breviarium Romanum, which is the normative text for this Project. Therefore this change will not currectly be accepted.

Issue #525, opened 2016-05-15
@ references in commemorations can loop infinitely


gregordick:
Pasc6-0 had this:

```
[Commemoration 3]
@:Commemoratio 2
```

Infinite loop ensued.

fiapps:
#3869 shows the importance of addressing this. The usual way to prevent an infinite loop is to count how many references have been followed. When this nesting depth exceeds some value, e.g., 10, the code returns an error like "References too deeply nested" instead of continuing to follow references.

FAJ-Munich:
Interesting issue; I've never understood what the original problem with the commemoratio3 was about. So thanks for explaining about the nesting issue. I guess the easiest way to avoid too deep nesting is to reference the actual text source as directly as possible, irrespective of the nesting in the breviary books?

fiapps:
> Interesting issue; I've never understood what the original problem with the commemoratio3 was about. So thanks for explaining about the nesting issue. I guess the easiest way to avoid too deep nesting is to reference the actual text source as directly as possible, irrespective of the nesting in the breviary books?

I would say that nesting should reflect the structure of the data in this repository. Since the repository does contain duplicated texts, it may not always be obvious which location is the definitive one to reference, so deep nesting could be created legitimately or at least in good faith. The problem isn't deep nesting, but circular nesting, and sometimes the circularity is not obvious.

When the issue was reported, `[Commemoratio 3]` was defined as `@:Commemoratio 2`, which in turn referenced `@Tempora/Pasc5-4:Oratio`. That didn't look circular, and I'm not certain what went wrong, but I think that when the code tried to resolve `@:Commemoratio 2`, it said, "oh, in this case I should be looking for Commemoratio 2 in the `[Commemoratio 3]` section." It did, found the section's content was `@:Commemoratio 2`, and tried to look that up, adjusting it yet again to `[Commemoratio 3]`, and so went round and round until someone killed the process.

[...]

PR #3866, opened 2024-06-14
Spanish translation and English simplification using @ references.

d4h0y0sr:


FAJ-Munich:
@d4h0y0sr  
As I see you are not making use of the inheritance of @ references: in most cases, these can go into the Latin and then they are not necessary in the English anymore as well. Of course, if Latin has the reference but English has the text, then the @ reference is not inherited to the other vernaculars like Spanish anymore. So maybe you can simplify even better by checking the Latin as well first. 

+Laudetur Jesus Christus+

FAJ-Munich:
"Now making use of" — sorry for the wrong autocorrect

d4h0y0sr:
Oh ok Thanks for the aclaration

APMarcello3:
@d4h0y0sr Please also be aware of issue #525 when simplifying texts.

PR #3869, opened 2024-06-16
Revert "refactoring using @reference inheritance"

APMarcello3:
Reverts DivinumOfficium/divinum-officium#3867

Causing infinite loops on server.


d4h0y0sr:
> Reverts #3867
> 
> Causing infinite loops on server.

I found out the issue, it was in file c10, i made a mistake and ended refering to itself.

PR #3889, opened 2024-06-20
perl syntax fix

gallizoltan:
fixing syntax error at horas.pl and specials.pl


gallizoltan:
@mbab What do you think about this?

APMarcello3:
Removes final antiphon of the BVM.

gallizoltan:
> Removes final antiphon of the BVM.

@APMarcello3 You are right, my mistake, I'm sorry for that!

Anyway, I have an old setup with an ancient Perl version, which is not compatible with this and is not easy to update. Could we please try another workaround? #3893

Issue #3921, opened 2024-07-03
Double "Deo gratias" at None on the feast of St. Rita


fiapps:
Yesterday, b0f82506ef7faea0c0dc13f8861a9fdb288ad8b5 added `$Deo gratias` to `[Capitulum Nona]` of C7a. As a result, There's now a double "℟. Deo grátias." after the capitulum in the Latin column of [None for St. Rita](https://www.divinumofficium.com/cgi-bin/horas/Pofficium.pl?date1=5-22-2025&command=prayNona&version=Rubrics%201960%20-%202020%20USA&testmode=regular&lang2=English&votive=).

APMarcello3:
[1130b05](https://github.com/DivinumOfficium/divinum-officium/commit/1130b0576d6926b4db3f30b4a327ddc4fec9ce17)

PR #3926, opened 2024-07-04
Commune C7ap: Double Deo gratias  #3921

FAJ-Munich:
@fiapps  Please let me know if there are other instances of doubled *Deo gratias* (potentially arising from circular references in Commune _in paschal tide?_)


fiapps:
> @fiapps Please let me know if there are other instances of doubled _Deo gratias_ (potentially arising from circular references in Commune _in paschal tide?_)

No other instances show up in my test cases.

PR #3934, opened 2024-07-07
Brute Force solution to #525 + Missa version names

FAJ-Munich:
1. breaking nested loops after six iterations 

@APMarcello3  @mbab @fiapps  : This should resolve the CPU overload; however, maybe the limit of 6 nestings is too narrow? Please test if any working references are broken by this limit.

2. Renaming the Missa version names hopefully resolving some other issue within the Cache for the `setupstring` function.


web/cgi-bin/DivinumOfficium/RunTimeOptions.pm
mbab, line 1: @FAJ-Munich  removing line with `1960 Newcalendar` causes error.

PR #3999, opened 2024-08-18
Revert "New Script File: HashParser.pl & Move of unused functions"

APMarcello3:


APMarcello3:
Undefined subroutine &main::date_to_days called at /home/apm/divinum-officium/web/cgi-bin/horas/horas.pl line 1045.

Issue #4082, opened 2024-10-07
Two Deo gratias for None of Our Lady on Saturday


rl1234567890:
I noticed that it is now:

Capitulum Responsorium Versus {ex Proprio de Tempore}
Sir 24:19-20
In platéis sicut cinnamómum et bálsamum aromatízans odórem dedi: quasi myrrha elécta, dedi suavitátem odóris.
℟. Deo grátias.
℟. Deo grátias.

PR #4083, opened 2024-10-07
Update C10.txt

FAJ-Munich:
weird behaviour of these references; not the first time for Capitulum Nona though and only in Latin.  #4082


mbab:
`$Deo gratias` could be added in code, to all Capitulums and ignored or removed in data
and `Capitulum Nona` could have defaults to `Lectio Prima` or inversely

PR #4096, opened 2024-10-14
code: modularize function special, split specials.pl

mbab:
+AMDG+

### DIFF
--- a/web/cgi-bin/horas/specials.pl
+++ b/web/cgi-bin/horas/specials.pl
@@ -7,6 +7,13 @@ $a = 4;
 
 use DivinumOfficium::Directorium qw(dirge);
 
+require "$Bin/specials/capitulis.pl";
+require "$Bin/specials/hymni.pl";
+require "$Bin/specials/orationes.pl";
+require "$Bin/specials/preces.pl";
+require "$Bin/specials/psalmi.pl";
+require "$Bin/specials/specprima.pl";
+
 #*** specials(\@s, $lang)
 # input the array of the script for hora, and the language
 # fills the content of the various chapters from the databases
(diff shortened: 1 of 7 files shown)


FAJ-Munich:
Great work to modularise specials. 

Thinking about issue #626 lately, I realised the functions that do perform formatting "too early" are those script functions mainly found in `horas.pl`. So I have been thinking about bringing all those called by the `&` directive into a single file, e.g. `horasscripts.pl`. As such, I wonder if Martyrology and Antiphonas finalis shouldn't become `#` sections covered by specials already? I guess this could have an impact on the epubgen so I'm hesitant as of now. What do you think, Father and @mbab ?

Additionally, I think those scripts called by missa as well (like do_io.pl , dialogcommon.pl , etc. ) should go into the DivinumOfficium folder?

mbab:
> As such, I wonder if Martyrology and Antiphonas finalis shouldn't become `#` sections covered by specials already?

As there is already `#Martyrology` in `Prima.txt` I bet You can try.

Maybe we should keep in ordinarium only `skeleton` and add middle level file (can't find proper name) with content as 
```
[De Officio Capituli]
$Deus in adjutorium iij
&Gloria
$MLitany2
$respice
&Gloria
$Oremus
$dirigere
$Amen
```

> Additionally, I think those scripts called by missa as well (like do_io.pl , dialogcommon.pl , etc. ) should go into the DivinumOfficium folder?

Certainly. And it doesn't look like it would be difficult to convert it into a perl modules

-----------------------

As for #626 I have idee:
change `specials` to ordinarium compiler which outputs something like:
```
[...
['Psalterium/Translate#Hymnus', sub { $_[0] =~ s/^/#/ }]
['Psa;terium/Minor Special#Hymnus Tertia', sub { # apply doxology, initial etc. }]
...]
```
generally 'source' of text from database + function which will perform necessary operation on it in next step,
output of such compilation can be linked for specified language to something like `markdown`, which can serve as base for HTML (and other formats) output.
This will make obsolete "vernacular" variables (as `%winner2`), `translate` and `prayer` functions and I believe should make code cleaner.
But I'm a bit worry about performance and size of work to implement this.

mbab:
I'm giving notification to this for @APMarcello3  & @FAJ-Munich as I doubt my reply reached you.

PR #4104, opened 2024-10-17
set commune in [Rule] as in [Rank]

mbab:
@FAJ-Munich wonder why we have Commune in two places

+AMDG+


FAJ-Munich:
I have no idea why we have it twice. But I can confirm the `[Rule]` version is used by the `getfromcommune()` function only which, again, is only invoked by the `getcommemoratio()` and maybe `lectio` routines. Didn't you just remove the former?

mbab:
getfromcommune() was removed in 77c7fb6ec9

PR #4151, opened 2024-10-30
quick hack to prevent perl warnings in FileIO.pm

mbab:
eg. Laudes 12/14/2024 1960

@FAJ-Munich `@commemoentries` contains empty strings, could fix this in proper place?

+AMDG+


APMarcello3:
Warning still occurring:

Screenshot_20241030_144834 1

Issue #4172, opened 2024-11-12
Introit of St. Martin Pope and Martyr


rl1234567890:
The Introit for the Mass of today is for paschal times:

Introitus
Ioann 21:15-17
Si díligis me, Simon Petre, pasce agnos meos, pasce oves meas. Allelúia, allelúia.

PR #4173, opened 2024-11-12
Dirty fix of Lectio94 S. Franciscæ Xaveriæ Cabrini

FAJ-Munich:
@mbab I think the variable `%w` in the `lectio : ScripFunc` is too overloaded and this script needs a logical overhaul to be streamlined and simplified


FAJ-Munich:
... and close #4172

PR #4181, opened 2024-11-20
allow translation of days ranks using data in Comment.txt

mbab:
Comment.txt in vernaculars needs to be filled.

+AMDG+


fiapps:
@mbab This doesn't matter too much since it's not visible in the browser, but there are now two newlines after the name of the celebration, as shown below. It matters to me because it means almost every test case has three changed lines, but I temporarily added `$dayhead =~ s/\s+/ /g;` to suppress the extra whitespace so I could test.

```
            <FONT>S. Felicis de Valois Confessoris

             ~ III. classis</FONT>
            <br/>
```

mbab:
fixed in #4182

PR #4206, opened 2024-12-14
pre '60: fix 12-14 readings

mbab:
+AMDG+


FAJ-Munich:
I think it should be covered by using the (Feria 4) tag in #4205 already. Sorry for not updating you via email, too.

mbab:
> I think it should be covered by using the (Feria 4) tag in #4205 already. Sorry for not updating you via email, too. +LJCh+
@FAJ-Munich Sure. For unknown reason I didn''t check if it is already done.
I didn't notice we have (feria 4).

PR #4224, opened 2024-12-21
update Kalendarium and Ranks

mbab:
+AMDG+


web/www/horas/Latin/Sancti/05-12.txt
fiapps, line 1: @mbab Merry Christmas! This change does not result in the correct rank for 1960 rubrics. This is a III class feast, but this change reduces it to a commemoration. See [2025-05-12 Laudes](https://www.divinumofficium.com/cgi-bin/horas/Pofficium.pl?date1=5-12-2025&command=prayLaudes&version=Rubrics%201960%20-%201960&testmode=regular&lang2=English&votive=), for example.

PR #4340, opened 2025-03-05
Revert recent accidental changes (plus an Italian fix)

fiapps:
- 4cbf012102a4e0b51acf76c01f25e0ad2e9195a4 added a second `Ant Vespera 3` section to `horas/Latin/Sancti/12-25.txt`. This shadows the section with the correct antiphons for the Roman Rite, so I'm reverting this change. I'm not sure what @Augustinus-Altovadensis was trying to accomplish. Perhaps it was meant to apply only to the Cistercian office. If it wasn't simply a mistake, he'll have to try again.
- dc500d9ec7d7ffffe3dabb65ceb1dde350413053 replaced three texts in `horas/English/Tempora/Epi2-0.txt` with Czech. I'm reverting this obvious mistake.
- There's also a fix for the Italian translation of Compline that I had forgotten to commit. This gives the oration Visita the same kind of initial capital in Italian that it has in Latin and other translations.

### DIFF
--- a/web/www/horas/English/Tempora/Epi2-0.txt
+++ b/web/www/horas/English/Tempora/Epi2-0.txt
@@ -5,7 +5,7 @@ Dominica II post Epiphaniam;;Semiduplex;;5
 God hath received * Israel his servant, as he spoke to our fathers, to Abraham and to his seed for ever.
 
 [Oratio]
-Všemohoucí, věčný Bože, jenž věci nebeské i pozemské řídíš, snažné prosby lidu svého milostivě vyslyš a mír svůj uděl našim časům. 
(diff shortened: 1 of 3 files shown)


Augustinus-Altovadensis:
Oh my, sorry for the mistakes!
- Yes, the second set of `Ant Vespera 3` starting with _In princípio._ was meant for the Cistercian version, and I forgot the magic comment. I'll do it again, this time with `(rubrica cisterciensis)`, as it was supposed to be.
- Yes, this is also the mistake. I usually copy the current English text as a basis for the Czech translation, because the old database files from my version don't always work. And I must have mistakingly open the file instead of copy&open... Thanks for correcting it!

PR #4417, opened 2025-04-10
1955: Invit in Passiontide

mbab:
+AMDG+


APMarcello3:
@mbab This invitatory antiphon is used along with the Pius XII psalms. It is not exclusive to 1955 rubrics. So, when Pius XII psalms are selected (for whatever rubrics), this antiphon should be used in Passiontide.

PR #4444, opened 2025-04-30
For issue #4443

### LINKED ISSUE #4443: Cist: in 31.5.2025, only Latin Comm. of Octave Sunday appear, Czech don't. And English do. Curious.
As the title says, in the Cistercian version de Altovado, in 31.5.2025 Vespers, only Latin Commemorations of Octave Sunday appear, Czech don't. However in English they do. Curious.

In Lauds, it shows (correctly) the Comm. of bl. Bernard, Mary and Grace, but the Sunday in Ascension Octave (2 Comm.) is still missing.

Now I've gone through all the languages, and they all work except for Czech, Spanish, French, Italian, and Hungarian. Strange.
And when I was working on this about a month ago, it worked. 

Any ideas? @FAJ-Munich, I suspect you might know better than most. 

Thank you, God bless!

FAJ-Munich:
close #4443

@Augustinus-Altovadensis 
Apparently, since in Latin, `Officium` and `Rank` is only defined in Tempora/Pasc6-0 but not TemporaM/Pasc6-0 there must not be a `Officium` hash in the vernacular versions of TemporaM/Pasc6-0. In Tempora/Pasc6-0 it is fine.

### DIFF
--- a/web/www/horas/Bohemice/Tempora/Pasc6-0.txt
+++ b/web/www/horas/Bohemice/Tempora/Pasc6-0.txt
@@ -4,12 +4,6 @@ Dominica infra Octavam Ascensionis
 [Officium] (rubrica 196 aut rubrica 1955)
 Dominica post Ascensionem
 
-[Rule]
-vide Tempora/Pasc5-4;
-9 lectiones
-Doxology=Asc
-StJamesRule=John,Rev
-
 [Ant 1]
 Až přijde * Utěšitel, kterého vám sešlu, co Ducha pravdy, který z Otce vychází, on bude vydávat svědectví o mě, alleluja.
 
--- a/web/www/horas/Bohemice/TemporaM/Pasc6-0.txt
+++ b/web/www/horas/Bohemice/TemporaM/Pasc6-0.txt
@@ -1,2 +1,0 @@
-[Officium]
-Dominica infra Octavam Ascensionis
--- a/web/www/horas/Espanol/TemporaM/Pasc6-0.txt
+++ b/web/www/horas/Espanol/TemporaM/Pasc6-0.txt
@@ -1,2 +1,0 @@
-[Officium]
-Dominica infra Octavam Ascensionis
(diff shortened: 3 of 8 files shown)


Augustinus-Altovadensis:
> @Augustinus-Altovadensis Apparently, since in Latin, `Officium` and `Rank` is only defined in Tempora/Pasc6-0 but not TemporaM/Pasc6-0 there must not be a `Officium` hash in the vernacular versions of TemporaM/Pasc6-0. In Tempora/Pasc6-0 it is fine.

Interesting! I'll try to remember that. So if I understand correctly, if in Latin those two hashes are not present, but in a vernacular yes, it will result in the program to ignore the original (`Sancti/Tempora`) file to a certain extent? 

And if these hashes are present in both, then the ones from `SanctiM/TemporaM` will be read and the ones from `Sancti/Tempora` will be ignored? 

BTW, is there any recommendation to do this? Like for me, it would make sense to keep those hashes on one place only, in order to avoid situations like this. Using a `[Rank] (rubrica monastica)` would effectively mean the same like placing it in the `SanctiM/TemporaM` folder, but without such adverse effects, correct? Or is there some recommended way already pointed out in the documentation and I just don't know about it?

FAJ-Munich:
I haven't figured out why but the situation is as follows:

If there is an `[Officium]` in a file, say `Bohemice/TemporaM/Pasc6-0` the algorithm looks for a `[Rank]` in the same file or, since we want them only in Latin, inherited from `Latin/TemporaM/Pasc6-0`. Since there was none, the recombined rank was `Dominica IV post Pascha;;` implying a numerical rank of zero. Hence no commemoration. 
For some reason it overwrites the inheritance from the preamble which says `@Tempora/Pasc6-0`. 

If the Czech translation of `[Officium]` is inherited via the preamble from `Bohemice/Tempora/Pasc6-0` the correct `[Rank]` is retrieved from `Latin/Tempora/Pasc6-0`.

Augustinus-Altovadensis:
Interesting. Makes sense in a way, and we can avoid that issue quite easily, it's more practical than trying to correct this behaviour and possible introduce other errors.

PR #4449, opened 2025-05-05
German translation (ordinary)

eiebehwiakwheh:
Added the German translation of "Go the Mass is ended" and "Thanks be to God"

image

Sourced from a German translation of the mass

https://archive.org/details/DasMessbuchDerHlKircheMissaleRomanumSchott1894/page/n75/mode/2up?view=theater


eiebehwiakwheh:
I'm afraid that my edit might have caused an error of some sort.

My intention was to remove the English and replace it with German.

Screenshot_20250507_174223

PR #4478, opened 2025-05-20
Cist: Upd. November + Czech

Augustinus-Altovadensis:
- In both Cistercian versions updated the month of November
- Czech translation added
- in Czech Psalms, the orthography of "Izrael" changed to older, but more correct "Israel"
- bug fixes


web/www/horas/Latin/Sancti/10-DU.txt
```diff
@@ -15,6 +15,11 @@ Ecce Vir Oriens * nomen ejus: sedébit et dominábitur, et loquétur pacem gént
 Dóminus * judex noster, Dóminus légifer noster: Dóminus Rex noster, ipse salvábit nos.
 Ecce dedi te * in lucem géntium, ut sis salus mea usque ad extrémum terræ.
 
+[Responsory Vespera 1] (rubrica cisterciensis)
```
mbab, line 18: @Augustinus-Altovadensis 
I wonder what is a reason to not use standard layout as:
```
R. Fecit ..
* Ipsi ...
V. Ipse ..
R. Ipsi...
```

Augustinus-Altovadensis: > @Augustinus-Altovadensis I wonder what is a reason to not use standard layout as:
> 
> ```
> R. Fecit ..
> * Ipsi ...
> V. Ipse ..
> R. Ipsi...
> ```

The reason is quite simple, in the Cistercian breviaries, it looks as I write it. I wanted to attach an image, but for some reason it doesn't work...

If it is an issue, I can do it according to the local standard, but to be honest, I hate to see an asterisk at the beginning of a line, it is just a very bad typography. But I don't know if it is of any use to argue about esthetics on DivinumOfficium, that looks pretty much the same since the 90's...

web/www/horas/Bohemice/Psalterium/Psalmorum/Psalm113.txt
```diff
@@ -1,5 +1,5 @@
-113:1 Když vyšel Izrael z Egypta, Jákobův dům z lidu temné řeči,
```
hason, line 1: @Augustinus-Altovadensis When modern Czech is used for translation, why is Izrael changed to Israel?

Augustinus-Altovadensis: @hason You know, the problem is that I argue with linguistic arguments and you with an appeal to authority. This approach happens to be considered a fallacy in some circles. Interesting that you list mainly New Testament translators and P. Václav Bogner, who was a generation or two younger. But that still doesn't make that argument valid from the professional point of view, which is all I'm interested in, and in the end, it is of course a personal view. 

And it has to be, as it is in the case of my translations of the Holy Office, and will be in some foreseeable future with the Psalms. Because as far as I know, there are no Czech translations of the Holy Office (apart from Tenebræ by P. Marian Schaller), and the last translation of the Scripture from Vulgate is more than a hundred years old, which makes it quite hard to understand for a modern Czech. 

It might not seem that way, but I am very accessible to constructive criticism, but only from the professional point of view. You see, I don't really care for authorities, especially those from the 60's and on, who resigned on the idea of having a truly Catholic biblical translation (cf. Lk 1, 28, words like ἐπίσκοπος in translations and many others). 

And as always: if you don't like my work, you are free and welcome to create your own translation, that will contain the exact right ratio of archaisms and linguistically incorrect modern spellings to your heart's content. But spitting on someone else's work that took a lot of study and effort doesn't really help anyone, does it?

PR #4488, opened 2025-05-24
Two different Czech translations

hason:
Originally, Schaller's Czech translation was added to this project. This translation is used in Holy Masses and is still being published.

@Augustinus-Altovadensis  is currently working on a new translation. I think it is a good idea to keep both translations and continue to incorporate them into the repository as contributors are able.


Augustinus-Altovadensis:
Given how "complete" is the implementation of Schaller's translation, the frequency of contributions, and the fact that he translated only the Holy Mass, I'd suggest to name the other translation after him. He is anyway _dignior mei,_ and he deserves the name more than I.

And my translations would remain simply `Bohemice`, as they are now.

If the admins deem it necessary, they could mention my name somewhere in the documentation page, but it's not really necessary.

hason:
It is consistent to use the shortened name for original and commonly used translations, which is Schaller (see Latin and Latin-Bea, Polish and Polski-Newer). I understand your excitement about creating your own translation, but that doesn't mean you will persevere and finish it. Or is this a task assigned to you by your superior, which guarantees that someone else will work on the translation after you?

Augustinus-Altovadensis:
I'm afraid you don't understand anything...
1) I don't want to have my name on the translation. I didn't write it in any of my books or leaflets published by the Monastery, and I don't want it here either, apart from some little comment in the documentation, if it **has to be.** Therefore, please, either create two folders: Bohemice-Emmer and Bohemice-Hason, or leave my name from this altogether.
2) I have already completed the translation. Now it's only a matter of copying it here. Maybe if you checked, you'd see that the Cistercian rite is complete in the Diurnale, and only the Matins are missing.

PR #4489, opened 2025-05-25
Czech: return to standard (Schaller kept)

Augustinus-Altovadensis:
I'm returning the Czech language to the state that it has been established at the beginning of its implementation.

@APMarcello3 Martin Hasoň (@hason) had no right to use my real name not only without my authorization, but against my expressed will. Please, never let him do that again. I do consider this incredibly rude and absolutely unacceptable. The translation has been put together by me, vastly expanded by me, but it's by no means _only_ my translation, therefore my name shouldn't just be slapped on it. 

Plus one important thing: DivinumOfficium is used also by elderly, who don't really know, how to change a language. Please don't make them look for help just for some bitter individual and stupid personal feud... Thank you.


Augustinus-Altovadensis:
@hason And I don't understand other things, yet here we are. 

Here is what I'll do. I will rename the folder and version carrying my name back to what it was. I have nothing against the existence of ten other Czech translations, if it is your wish. So the folder `Bohemice-Schaller` will remain, I will even rename it to your preferred name "Cesky-Schaller", as the code cannot handle the coexistence of `Bohemice` and `Bohemice-Schaller` correctly, and I don't want to spend any time on finding why... And you can work on it without any kind of interference from my side. Of course you are free to rename it, just make sure that both Czech versions are working properly. But please, never, NEVER attach my name anywhere! 

This is my last public comment about all of this, as I will rather spend my time working on the project than writing comments that fall on deaf ears.

PR #4572, opened 2025-06-24
MONASTIC 1617 - 12 readings: 06-25

mbab:
+AMDG+

### DIFF
--- a/web/www/horas/Latin/SanctiM/06-25.txt
+++ b/web/www/horas/Latin/SanctiM/06-25.txt
@@ -3,6 +3,21 @@
 [Name]
 Guliélmi
 
+[Rule]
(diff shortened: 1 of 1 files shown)


Augustinus-Altovadensis (commented):
@mbab Could you please limit this to other versions than the Cistercian one? The Breviarium Cisterciense clearly states _Laudémus viros gloriósos._ for the first Nocturn of St. Wilhelm, but this is _Justus si morte præoccupátus fúerit._, so you are introducing an error into the Cistercian version.

Please, when introducing new readings or other things like this, either comment the Cistercian version out, or check the Breviarium Cisterciense, which is freely accessible on Google Books. Thank you!

PR #4589, opened 2025-07-02
M1930 and Cist: fix S. Mauri (non-transferred)

FAJ-Munich:
@mbab 
the `transfered()` sub in `directorium.pm` got confused with the files sharing the date desgination `01-15` with and without `cc`.
Please note, the 12 lections for S. Mauri need to go in `SanctiM/01-15cc.txt`


mbab:
@FAJ-Munich  Can't we use `01-15` instead of `01-15cc`? if not will be correct to have 1617 version called `01-15t`?

FAJ-Munich:
> @FAJ-Munich Can't we use `01-15` instead of `01-15cc`? if not will be correct to have 1617 version called `01-15t`?

In the Roman, January 15 is S. Pauli primi Eremitae (in file `01-15`) with commemoration of S. Mauri in file `01-15cc`. Using the same file names for the Monastic ensures correct fallback to the Roman files when things are missing. It is a bit of a mess for the Monastic, as S. Pauli used to be on Jan 19 in M1617 then Jan 10 in M1930.

mbab:
> > @FAJ-Munich Can't we use `01-15` instead of `01-15cc`? if not will be correct to have 1617 version called `01-15t`?
> 
> In the Roman, January 15 is S. Pauli primi Eremitae (in file `01-15`) with commemoration of S. Mauri in file `01-15cc`. Using the same file names for the Monastic ensures correct fallback to the Roman files when things are missing. It is a bit of a mess for the Monastic, as S. Pauli used to be on Jan 19 in M1617 then Jan 10 in M1930.
Do we need fallback if Latin file is in place? 
How the file for 1617 should be called?

FAJ-Munich:
> Do we need fallback if Latin file is in place?

For ease of maintanance of the Transfer files, at least we should have consistency, i.e., the same file name should be used for S. Mauri also in the Roman. I used the `cc` designation as S. Mauri has not been standalone after 1570 at least. I guess `o` or `oM` is also fine–doesn't really matter. The issue, `directorium.pm` ran into did not arise from the appended letters to S. Mauri but the fact that S. Pauli in the file `01-15` is not used on that day at all in the Monastic. 

> How the file for 1617 should be called?

As S. Mauri is Duplex II. classis, isn't it a file for both 1617 and 1930 you are preparing?

mbab:
> As S. Mauri is Duplex II. classis, isn't it a file for both 1617 and 1930 you are preparing?

1930 is full office, 1617 all from Commune apart readings 5-8,

mbab:
> I see! Well, I guess you have two options then:

I will keep first option. Adding `(rubrica divino)` prevents from data being reusable & looks ugly.

PR #4618, opened 2025-07-13
Cist: bugfixes

Augustinus-Altovadensis:
Correction of typos and missed things.

@FAJ-Munich 
Could you please check, whether on Saturday January 11 in any monastic rite, the day within Epiphany Octave should have second Vespers? I trust you still have the rubrics fresh in mind... I checked the last week's changes and found that in Vespers on 1-11-2025 for the Altovadense version, the Commemoration of the Octave appears twice, and once (from the `SanctiM/01-11.txt` file) without the translation. This PR correct this, but it might introduce another problem...

BTW, for this day in the 1951 version, it doesn't show the commemoration of the Octave, which is in the `TemporaM?` file, which it should... In Vespers nor in Lauds.

Thank you, God bless!

### DIFF
--- a/web/www/horas/Bohemice/SanctiM/03-21C.txt
+++ b/web/www/horas/Bohemice/SanctiM/03-21C.txt
@@ -49,7 +49,7 @@ Duchu vznešený, jeden Bože, po vše
 Amen.
 
 [Ant 1]
-Přesvatý * Vyznavači Páně, otče a vůcče Mnichů, Benedikte, přimlouvej se za naši spásu i za spásu všech.
+Přesvatý * Vyznavači Páně, otče a vůdče Mnichů, Benedikte, přimlouvej se za naši spásu i za spásu všech.
 
 [Oratio]
 Vzbuď, Pane, ve své Církvi Ducha, jemuž svatý Benedikt, Opat, sloužil: abychom, jím také naplněni, snažili se milovat, co on miloval, a skutky vykonávat to, co vyučoval.
@@ -74,7 +74,7 @@ Vznešený řádů našich Patriarcha,
 Isáka, Mojžíše i Abraháma, 
 V srdci svém spojil.
 _
-On sám ty, které světa bouře smetly,
+On sám ty, které bouře světa smetly,
 Zde zbožným pláčem k životu navrátil, 
 Mír pak z místa, kde strach býti nemůže,
 V pokoji přines'.
@@ -87,7 +87,7 @@ Amen.
 
 [Ant Matutinum] (nisi tempore paschali)
 Byl to muž ctihodného života, * milostí požehnaný <i>(Benedictus)</i> i jménem.;;1
-Od samého svého dětství * měl srdce starce, svůj věk převyšoval svým chováním, a nepodvolil svého ducha žádným potěšením.;;2
+Od samého svého dětství * měl srdce zkušené, svůj věk převyšoval svým chováním, a nepodvolil svého ducha žádným potěšením.;;2
 Dokud byl ještě na této zemi, * i na to, co ještě svobodně mohl používat, již shlížel jako na vyprahlý svět s květy.;;4
 Vzešel * z provincie Nursie, a v Římě, kam jej rodiče poslali, obdržel klasické vzdělání.;;5
 Opustil dům * i majetek otce, a v touze zalíbit se pouze Bohu, hledal hábit svatého obracení.;;8
--- a/web/www/horas/Latin/Sancti/01-15.txt
+++ b/web/www/horas/Latin/Sancti/01-15.txt
@@ -5,7 +5,7 @@ S. Pauli Primi Eremitæ et Confessoris
 ;;Duplex;;3;;vide C5
 (sed rubrica 1570 aut rubrica 1617)
 ;;Semiduplex;;2.2;;vide C5
-(sed rubrica 1930 aut rubrica 1963)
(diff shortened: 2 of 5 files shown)


FAJ-Munich:
In general: yes the Monastic versions that do not celebrate the Feast of the Holy Family, the commemoration coming from the file SanctiM/01-11 needs to be there. It's treated similarly to the situation of the Corpus Christi Octave. I'd be surprised if the same doesn't hold for the Cistercian rite which loves commemorations. 

For Cist1951 on the Sunday:
This is a rare case which is added in the code. I guess `occurrence()` received a `&& $version !~ /cist/i` too much. 

For Altovado:
I'm surprised to see the doubled commemoration in vespers: either way, the one from the Feast should be removed and the one from the 01-11 retained for consistency of the code and database with the other versions.

Augustinus-Altovadensis:
> In general: yes the Monastic versions that do not celebrate the Feast of the Holy Family, the commemoration coming from the file SanctiM/01-11 needs to be there. It's treated similarly to the situation of the Corpus Christi Octave. 

Very good, so the best solution of this is to remove the commemoration from the Sunday's datafile, remove all those conditions from 01-11.txt and fix the rank for the Cistercian versions, which is ridiculously high. 

And also there is always some problem with the Czech version of a day within Octave, there's probably another place in the code that scans for `/Octav/i` in the Officium/Rank line, which wouldn't work for Czech, that calls it "Oktáv"...

> I'd be surprised if the same doesn't hold for the Cistercian rite which loves commemorations.

Of course, the commemoration is always there, so on the Feast of Holy Family, there is a commemoration of the Sunday itself and then the commemoration of the Octave. 

> For Cist1951 on the Sunday: This is a rare case which is added in the code. I guess `occurrence()` received a `&& $version !~ /cist/i` too much.

OK, but this is no exception, the Sunday is the main Office, and the Octave is always commemorated.

> For Altovado: I'm surprised to see the doubled commemoration in vespers: either way, the one from the Feast should be removed and the one from the 01-11 retained for consistency of the code and database with the other versions.

Agreed.

FAJ-Munich:
> Very good, so the best solution of this is to remove the commemoration from the Sunday's datafile, remove all those conditions from 01-11.txt and fix the rank for the Cistercian versions, which is ridiculously high. 

You mean the commemoration in the feast's file? Cist51 with the Sunday's file has been fine. 

Regarding the numerical ranks, it is okay and necessary to stay as high as it is: the situation is comparable to the ferias in lent and Holy Week: the Epiphany octave has always had very special privileges making their numerical rank comparable to a Serm. Maj. despite being iij. Lect. et M. 
The display of the ranks is corrected in the setrank function and works across the different versions including Cistercian and orders of privilidged Octaves: please compare the description in technical.html again. 

> And also there is always some problem with the Czech version of a day within Octave, there's probably another place in the code that scans for `/Octav/i` in the Officium/Rank line, which wouldn't work for Czech, that calls it "Oktáv"...

There is a variable called `$octavestring` which might need an adjustment to identify the Czech name correctly. Could be the source of it is hidden in the translate.txt.


> Of course, the commemoration is always there, so on the Feast of Holy Family, there is a commemoration of the Sunday itself and then the commemoration of the Octave. 
> 

Agreed. So we need to model it from the Roman DA version. Let me rename the existing SanctiM/Epi1-0.txt to match the Dancti folder (including all necessary adjustments) and then we can put Holy Family adjustments for Altovado in its proper place in 
SanctiM. This should also solve the issue with missing commemoration in Lauds of the Sunday under 1951.

Augustinus-Altovadensis:
> You mean the commemoration in the feast's file? Cist51 with the Sunday's file has been fine.

Actually no, the Sunday within the Epiphany Octave is missing all the commemorations of the Octave, which is wrong. On Sunday, the winner should be the Sunday within Octave, but the Commemoration of the Octave should be there as well, but currently is not.
 
> Regarding the numerical ranks, it is okay and necessary to stay as high as it is: the situation is comparable to the ferias in lent and Holy Week: the Epiphany octave has always had very special privileges making their numerical rank comparable to a Serm. Maj. despite being iij. Lect. et M. The display of the ranks is corrected in the setrank function and works across the different versions including Cistercian and orders of privileged Octaves: please compare the description in technical.html again.

Very good. In that case, the rank needs to be adjusted for both Cistercian versions, so that it falls under the Sunday. The funny thing is that currently, if the rank is 5.6, the commemoration in both Latin and Czech is displayed, and if I lower the rank, only the commemoration in Latin is displayed, but on its correct place below the Sunday.

> There is a variable called `$octavestring` which might need an adjustment to identify the Czech name correctly. Could be the source of it is hidden in the translate.txt.

Yes, I saw that and it's set up correctly, so that it catches the Czech version too. The problem seems to be somewhere else.

> Agreed. So we need to model it from the Roman DA version. Let me rename the existing TemporaM/Epi1-0.txt to match the Sancti folder (including all necessary adjustments) and then we can put Holy Family adjustments for Altovado in its proper place in TemporaM. This should also solve the issue with missing commemoration in Lauds of the Sunday under 1951.

Very good. I will gladly leave this up to you, you will know better, what to do. 
So to sum that up, on Vespers on 1-11-2025, if you could do:
- display the _Commemoration of Sexta die infra Oct. Epiph._ in both Cistercian versions
- in Altovadensis version we need to see first the Commemoration of the Sunday and then the Octave day
- _in both languages_.

The same goes for Lauds on 1-12-2025, the Vespers will be from Baptisma, so this will be a different story.

PR #4637, opened 2025-07-18
Cist: commemorations

FAJ-Munich:
@Augustinus-Altovadensis 
39f575bd2d59d7a5da6e60df2228228a95ace35f addresses your request.

However, I think a general question on the order of commemorations needs to be answered by the Brothers (or you):
- In the Tridentine rubrics and, hence, the Cistercian 1951, the order of commemorations purely follows the ranks of feasts and days. 
- Divino Afflatu changed this general rule in two ways:
1. Sunday commemorations are privilidged trumping all feasts (this is what interferred on Jan 12, 2025 in Altovado implementation)
2. The concurrent feast gets commemorated first always whatsoever. (This is not implemented for Altovado)

In the commit, I now relaxed rule 1 for the case of Sunday commemmorations comming out of `[Commemoratio]` hashes. The other option would have been to implement rule 2 for Altovado. The latter, however, would impact other days throughout the year also.

### DIFF
--- a/web/cgi-bin/horas/specials/orationes.pl
+++ b/web/cgi-bin/horas/specials/orationes.pl
@@ -271,10 +271,11 @@ sub oratio {
           $ccind++;
           my $key = ($ic =~ /$sundaystring/i)
             ? (
-              $version !~ /trident(?!.*Altovadensis)/i
-              ? 3000
-              : 7100
-            )    # Sundays are all privilegde commemorations under DA
+              $version !~ /trident/i
+              ? 3000                                  # Sundays are all privilegde commemorations under DA
+              : $version !~ /altovadensis/i ? 7100    # Dom. min.
+              : 6100                                  # Sundays are below MM.maj. commemorations for Altovado
+            )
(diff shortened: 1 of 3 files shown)


Augustinus-Altovadensis:
@FAJ-Munich 
Thank you!

Actually, this rule hasn't been implemented into the Cistercian Office.

In 1951, an ordinary Sunday is simply considered as a feast placed between xij. Lect. et M. and MM. min. No special treatment whatsoever. This of course changes for first class Sundays, which usually get the rank of Serm. maj. 

In today's Altovadensis Office, the only thing that has changed is the rank of Sunday, which has been moved between MM. min. and MM. maj. This has been done in accordance with the 1957 implementation of _Cum nostra hac ætate._ However, other provisions of this decree, like expanding the number of first class Sundays, have not been definitely introduced. 

So the correct way for the Cistercian rite is to treat an ordinary Sunday as every other feast with a rank of 2.5 for 1951 and 3.5 for Altovadensis. First class Sundays would then be the rank of 6.5, probably, there it depends a lot on the feast they are trumping...

PR #4673, opened 2025-08-13
Cist: Start Disentanglement

FAJ-Munich:
@Augustinus-Altovadensis  
As disscussed in #4372 this gives you the possibility to disentangle the Cistercian version from the Benedictine without changing anything that is not working. I've tested it for C1951 and CAV and did not come across any changes until the end of 2026. 
I think the next logical step would be to disentangle the Commune files but I won't have the time for now. Once a file is present in SanctiCist or CommuneCist or TemporaCist, it hides the corresponding SanctiM file etc.


Augustinus-Altovadensis:
@FAJ-Munich 
Oh my, this took quite a lot of work! Very good, when I manage to find some time, I will continue using the new Cist folders whenever warranted. 

Yes, the Commune makes sense.

I was also thinking about the Responsories on Sunday Matins, that created a bit complicated situations, but without changing all the Sundays' datafiles, this won't be possible.

PR #4699, opened 2025-08-30
Update Prayers.txt

abi-ramia:
The word bênção is preferentially spelled with a ^, as in Translate.txt. Currently, bênção is not in red in the Incipit in Portuguese because of this discrepancy.

### DIFF
--- a/web/www/horas/Portugues/Psalterium/Common/Prayers.txt
+++ b/web/www/horas/Portugues/Psalterium/Common/Prayers.txt
@@ -207,7 +207,7 @@ V. Da porta do inferno.
 R. Libera, ó Senhor, suas almas!
 
 [Oratio Visita_]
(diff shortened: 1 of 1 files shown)


abi-ramia:
Dear all,
Upon comparing the English and Portuguese versions of web/www/horas/"Language"/Psalterium/Common/Prayers.txt, I realized that probably the problem with the original version was not only the misspelling of "bênção", but also the usage of Portuguese instead of Latin for the beginning of the benediction (or at least this is the discrepancy I have now found).

Is this indeed the case? I.e., does the translation go from Latin → Portuguese and then apply the correct font?

Would then writing "Benedictio. Abençoe e nos guarde o Senhor onipotente e misericordioso, + Pai, e Filho e Espírito Santo." solve the issue?

Thank you very much for your help and patience. I'm trying not to ask too many questions while still grasping how to contribute, but this is the first time the issue I've encountered is not simply a type.

All the best!

FAJ-Munich:
I think it should work the way you have described. Benedictio gets translated using the spelling provided in translate.txt on the fly and then it should also be in red.

PR #4784, opened 2025-10-28
Monastic 1617 / 1930; Fix Compline and add rubrics

FAJ-Munich:
- Fixed Compline when Preces Dominicales are not to be said
- corrected the conclusion of the minor hours (no Divinum auxilium)
- added rubric before final Pater noster

The rubrics pointed out for Monastic 1930 are actually the same as in the Tridentine books. DO, however, presumes that the recitation is done in private and that there is a break after Lauds: therefore, the only two places to show the Final Antiphone are Lauds and Compline. — If someone is to join Prime and more hours after Matins and lauds, technically, he should postpone the Antiphone until he stops—but then that is "expert mode" which goes beyond the purpose of the website IMHO.

@mbab Please double check if I have overlooked something. I think just today's Feast of the Apostles is sufficient to have a look at all the hours conclusions.

### DIFF
--- a/web/www/horas/Ordinarium/Completorium.txt
+++ b/web/www/horas/Ordinarium/Completorium.txt
@@ -45,7 +45,7 @@ $Deus in adjutorium
 (sed rubrica 196 aut rubrica 1955 aut rubrica cisterciensis omittitur)
 
 #Oratio
-(si rubrica ^Monastic aut rubrica cisterciensis dicuntur)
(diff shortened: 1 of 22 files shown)


web/www/horas/Ordinarium/Completorium.txt
FAJ-Munich: This is going to fix the issue with Preces. (copied from the solution for Prime)

PR #4820, opened 2025-12-03
Add Vietnamese translation for Office and Mass

blablapro:


FAJ-Munich:
@blablapro  
In the vernacular, no redundancies should be added, i.e.
- files that are purely links (e.g., Commune/C3bp.txt) should not be included in the vernacular at all
- hashes which are just links, or should not be overwritten from Latin, (e.g., `[Rule]` or `[Rank]` if `[Officium]` is also present) should also not be included in the vernacular files.

As a general rule, if the file or hash is not present in English, it certainly should also not be present in any other vernacular.

Please reopen this PR after kindly removing these redundant files and sections.

PR #4887, opened 2025-12-29
Magyar: Hungarian names of saints in horas

vlacko0930:


web/www/horas/Magyar/Sancti/09-29.txt
vlacko0930: Too much deletion

PR #4901, opened 2026-01-01
CIST: Completed and fixed Vigil and Feast of Epiphany + fixes

Augustinus-Altovadensis:
Fixes:
- removed Suffragium for the Vigil of Epiphany of Our Lord
- in the Cistercian rite, the festive Hymns never change the doxology, only ferial. Fixed.
- in 1951 Cistercian Breviary, the feast of the Most Holy Name of Jesus comes on the second Sunday after Epiphany. Fixed.


Augustinus-Altovadensis:
> The Tridentine also has Most Holy Name on 2nd Sunday after Epiphany (or Jan 28 if Easter is too early). You'll find the transfers using the Sanctoral file 01-00 for these cases.

Yes, it was already filled in correctly for C1951 before, so I just removed the useless duplication that I might have introduced in the past...

I guess putting the Most Holy Name on the _dominica vacans_ in the first week of January is a direct - and quite logical - consequence of removing the Octave Days of St. Stephen, John, Holy Innocents and the Epiphany Vigil in 1955.

PR #4902, opened 2026-01-01
Language of calendar

vlacko0930:
Add vernacular to calendar

### DIFF
--- a/web/cgi-bin/horas/horascommon.pl
+++ b/web/cgi-bin/horas/horascommon.pl
@@ -127,11 +127,11 @@ sub occurrence {
 
       if ($tomorrow) {
         $tvesp = 1;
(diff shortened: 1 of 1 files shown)


FAJ-Munich:
Changing this in horascommon could result in critical problems with the code as many checks presume the ranks and officium names to be in Latin! I doubt the results would be accurate if lang1 does not equal Latin otherwise. 

If you want the calendar output itself in the vernacular, necessary additions would need to be placed in the scripts of kalendar.pl only. 

I vaguely remember that their is a possibility to have the Ordo in the vernacular. Maybe @fiapps or @mbab remember/know?

mbab:
If You want vernacular in *Ordo* there is not need to change code, just set lang1 (Language for left column) in Options.
But if it is about *Kalendarium*, some changes in `kal.pl` will be needed.

PR #5041, opened 2026-02-22
Create test_env.psgi

tjrandall:
Wrapper file to run Divinum Officium using Plack, without having to rewrite any of the CGI functionality.  I tested both on Termux, and my local environment, and the speed / functionality seemed normal.  I did not load test to see how running in Plack changed performance.

Note that I added the commands I needed to run to install dependencies in Termux.

### DIFF
--- a/test_env.psgi
+++ b/test_env.psgi
@@ -1,0 +1,65 @@
+use Plack::Builder;
+use Plack::App::CGIBin;
+use Plack::App::File;
+use Cwd qw(abs_path);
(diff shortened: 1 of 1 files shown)


tjrandall:
Will need to continue to load test to verify if Plack can improve performance.

APMarcello3:
@tjrandall Before I merge the PR, could you change the instructions in download.html for > Hosting Locally: Python < (perhaps change to Plack)?

PR #5125, opened 2026-04-16
adding a robots.txt file to block crawlers

tjrandall:
Adding robots.txt to app.psgi ensures blocking bot requests

### DIFF
--- a/app.psgi
+++ b/app.psgi
@@ -72,6 +72,14 @@ builder {
                 "User-agent: PerplexityBot\n" .
                 "Disallow: /\n\n" .
 
+                # Block Ahrefs (SEO Crawler)
+                "User-agent: AhrefsBot\n" .
+                "Disallow: /\n\n" .
+
+                # Block Barkrowler (SEO Crawler)
+                "User-agent: Barkrowler\n" .
+                "Disallow: /\n\n" .
+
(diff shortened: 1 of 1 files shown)


fiapps:
If the goal is to reduce server load, wouldn't it make more sense to add a static `robots.txt` file to `web/`, where the root handler can serve it, instead of producing it as dynamic content?

tjrandall:
@fiapps - I know that serving a static file is standard, but I opted for the dynamic mount in app.psgi because the text is in-memory, so it's served faster than a file-system look-up.

If we want to move to a static file, I'm fine with that as well.

fiapps:
@tjrandall My gut feeling is that the speedup in practice would be negligible because the limiting factor is network operations. `robots.txt` would probably be cached in memory most of the time you access it, so you wouldn't have a cost for reading from disk. It would just be the system call overhead. But that overhead is measured in microseconds, and network communication times are measured in milliseconds.

I also asked Claude Opus, and it agrees this is not a meaningful optimization in practice. It adds, "`Plack::Middleware::Static` or `Plack::App::File` will serve a static `robots.txt` efficiently. `Plack::App::File` even sets `Last-Modified` and handles `If-Modified-Since` for you, so well-behaved crawlers won't even need the body most of the time — they'll get a `304 Not Modified.`" The dynamic `robots.txt` doesn't have that optimization.

PR #5153, opened 2026-04-30
Fix Plack 405 error by pinning Plack to 1.0050

tjrandall:
- Pin Plack to 1.0050 to avoid return_405 method mismatch in Plack::App::File
- Add arch-aware dumb-init download (x86_64 + aarch64)
- Move PERL5LIB setup to startup time instead of per-request
- Remove unused Cwd import and curl dependency
- Clear debug env vars (PERL5OPT, PERL5DB, PERLDB_OPTS) at container start
- Guard unequivocal() in RunTimeOptions.pm against undefined value argument


tjrandall:
Updates the recent Plack 405 errors.

PR #5170, opened 2026-05-06
Speed up Ordo

FAJ-Munich:
@tjrandall 
Please check if these shortcuts are sufficient to make caching superflous.


tjrandall:
this is a great PR @FAJ-Munich - should ensure that the Totus caching is much faster.

PR #5188, opened 2026-05-13
Add word-by-word interlinear Latin gloss option

rtpmk3:
Adds an optional interlinear display that shows short English translations under each Latin word. Enable it in Options under the Interlinear setting.

Clicking a word hides its gloss; clicking again restores it.

Some glosses may be missing or incorrect — corrections go in lexicon_overrides.json.


APMarcello3:
@FAJ-Munich @mbab Kindly review?

FAJ-Munich:
- First glance seems okay. I haven't tested with Docker though, just Apache for now.
- It should be ensured that the option is not activated as per default.
- @tjrandall  Could you run some performance tests. I am not sure what it means to send the json-Files with 80000+ entries per request.

tjrandall:
@FAJ-Munich  — good catch on the JSON concern. Before I run tests, I want to make sure I'm measuring the right things. My plan is to test three scenarios: baseline on current master, the interlinear branch with the option OFF, and the interlinear branch with the option ON. For each I'll capture response time and payload size.

The key questions I want to answer are: 

1. is the lexicon JSON loaded per request or once at startup
2. What's the actual file size in practice, and 
3. does it get sent to the client even when the feature is disabled. T
 
That last one is especially important given your note about the default setting.

I'll start with local testing first to understand the behavior before pushing anything to Cloud Run. Will report back with numbers.

tjrandall:
@FAJ-Munich — completed local Docker testing. Here's what I found.

**Code review first:** 
The implementation is clean - great work! 

`apply_interlinear` is gated behind the `$interlinear` flag in webdia.pl — it never runs when the feature is off. The lexicon loads once per worker process on first use (the `$_loaded` flag handles this), not per request. So users who never enable interlinear pay zero cost.

**Default setting:** Confirmed `$interlinear='0'` in horas.setup ✅

**Performance numbers:**

| | Payload | Mean Response |
|---|---|---|
| Interlinear OFF | 19,317 bytes | ~430ms |
| Interlinear ON | 132,776 bytes | slower* |

*The browser load time with interlinear ON was noticeably slower, primarily driven by payload size — **6.9x larger** than baseline. Server-side processing adds some overhead on first request per worker (lexicon parse of ~2.5MB JSON), but subsequent requests benefit from the in-memory cache.

**Honest concern:** The payload size is the real issue for users on slow connections. 132KB vs 19KB is significant. That said, this is an opt-in feature and the server-side implementation is efficient.

No blocking issues found. Suggest merging with a note in the UI that interlinear may affect load times on slower connections.

rtpmk3:
I noticed occasional cloudflare timeouts on divinumofficium's site with interlinear enabled, but they resolve after reloading the page. I will create a new branch to investigate making this more efficient and will submit another PR if I am happy with it. I am also happy to make translation corrections as I or other users come across them.

APMarcello3:
@rtpmk3 Please see issue #5192

rtpmk3:
> @rtpmk3 Please see issue #5192

Ok I will look that this.

PR #5206, opened 2026-05-15
Fix interlinear appearing enabled for users with old cookies

rtpmk3:
Users with cookies saved before interlinear was added had the old `parameterscheck` string land in the `$interlinear` parameter slot when loaded. Since that string is non-empty, it was interpreted as "on" — so interlinear appeared enabled even though the user had never selected it.

Fix: reset `$interlinear` to 0 on page load unless explicitly set via the Options dialog or a URL parameter. Applied to `officium.pl`, `missa.pl`, `Cmissa.pl`, `popup.pl`, and `mpopup.pl`. Also corrects the `parameterscheck` length in `horas.setup` and `missa.setup` after `$glossfont` was added.

Testing: Reproduced by manually setting old pre-interlinear cookie values (`horasp` and `missap`) in the browser and navigating directly to the horas and missa pages. Confirmed the bug on the live site, and confirmed the fix works against the docker container instance with this fix.

### DIFF
--- a/web/cgi-bin/horas/officium.pl
+++ b/web/cgi-bin/horas/officium.pl
@@ -119,6 +119,8 @@ if (!$setupsave) {
 set_runtime_options('general' . ($Ck ? 'c' : ''));    #$expand, $version, $lang2
 set_runtime_options('parameters');                    # priest, lang1 ... etc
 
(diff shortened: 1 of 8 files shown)


rtpmk3:
I apologize for the inconvenience

PR #5419, opened 2026-08-05
Implement generative martyrology based on elogia pool; partly address #4412

pythonengineer:
This is a bulky pull request that effectively rewrites everything, including both the data and how it's processed, pertaining to the Martyrologium: we completely overhaul the martyrology raw month-day files into carefully converted elogia data files, still month-day but now capable of dynamic inheritance and reordering based on what translation and version we're on. This means that the original month-day files will now be there just for reference in the `source` folder.

The point of this is to encourage contributions of any scale to the various versions of the martyrology without having to worry about violating the principle of DRY because we're copying an entire raw data file into a new folder just to alter, add or subtract a single elogium. Before you would have to do that, as of this change you'd still need the folder and the file but now that file could be 2 lines and *only* the pure substance pertaining to that year and translation.

Of course, future efforts to add a new translation would now require a conversion process: that is the purpose of the new `martyrologium-tools` directory. All existing translations and versions from the Latin are covered already by this pull request, and the new tool used to verify the conversion of such as the same on-screen result as the source files in use now shows that full equivalency is reached, which means that the "keys" that describe who an elogium is for and where it goes are all already defined, making adding a new translation as easy as using the existing `split-months` tool to get the type of raw files we've already been using and using the new `import_translation` Python tool to generate the pool of elogia files out of them. Matching who goes where according to the Latin as a schema is designed to be automated and has worked for the existing languages.

While none of this visibly affects anything on-screen just as of this pull request, the power of this system is shown in the new off-by-default "left column elogia as fallback" option: the server is now able to easily render an elogium that is missing from the right-side translation using the left-side language because of the inheritance structure of language and version that is now recognized:

toenglish

Same can be done if you, for instance, put English on the left side and French on the right side, a key only present in English will appear in English on the French side. This is more of a debugging tool (though I personally enjoy it) but it proves that it's able to match the elogia across languages.

This of course doesn't specifically resolve any longstanding issues of certain translations not matching the exact 1960 Latin or what have you, elogia missing whichever way, but it clears the way for making fixing those issues *way* more manageable.


mbab:
Maybe instead of option it should default to 'Fallback language' unconditionally?

On other side: do we really need another programming language?

pythonengineer:
> Maybe instead of option it should default to 'Fallback language' unconditionally?

I set it to off by default because I wasn't sure if straight up inserting a different language elogium (even if it's the language the user chose for the left side) would be a popular preference, but if that's agreed upon to be the more appropriate move if the right side is missing elogia, sure.

> On other side: do we really need another programming language?

Not need per se. Those scripts are standalone for converting new translations in bulk to this format

APMarcello3:
@mbab @FAJ-Munich kindly fully review and advise if and when this PR will be ready for merge.

mbab:
:+1: 
the only question is if it should be option, or just 'use' of **fallback** language (not **left side** certainly),
as we fallback in other places.

FAJ-Munich:
I'll have look later tonight.

fiapps:
Something like this has long been needed. "The translation of the Martyrology doesn't match the Latin" is a common complaint, and storing translations by *elogium* seems the most practical way to resolve this.

I haven't evaluated the solution in detail, but it looks good. I have just one suggestion: put the offline tools for processing Martyrology files under `admin/`, where other such tools are kept.

FAJ-Munich:
First few observations:

- agreed with @fiapps :  the tools should go into `/admin`
- similarly, the old source files should go into `/obsolete`
- agreed with @mbab :  if possible, the phython script should be converted into perl. from experience with DO it should be fairly simple and with AI help probably even easier...   thereby, the tools can actually be used by everybody
- structure of files:
 1. The files should use the current system of DO, i.e., setupstring functionality. 
 2. Thereby, we should only have one folder `Martyrologium` in each Language folder and the other folders become obsolete.
 3. There are hash names in the files which are arbitrary like `[In-Africa]` in 01-01.txt
 4. There are differences left in other versions which only stem from typos punctuation differences (e.g., [Almachii] in 1570 vs. DA 01-01.txt or [Circumcisio-Octava] DA vs. 1955R)
 5. 
 6. There should be a hash `[Martyrologium]` in each Latin file, and Latin file only, which composes the full text from the other hashes. Thus, the question of *fallback* languages or missing entries or multiple versions can be dealt with unambigously:

For instance, for January 1:
```
[Titulus]
Kaléndis Januárii

[Martyrologium]
@:Circumcisio-Octava
@:Martinae
@:Basilii
@:Bonfilii
@:Almachii
@:Item-Romae
@:Concordii
@:Magni
@:In-Africa
@:Justini
@:Eugendi
@:Odilonis-Cluniacensis
(deinde dicitur)
@:Vincentii-Strambi
(sed rubrica 1570 omittitur)
@:Euphrosynae

[Circumcisio-Octava]
Circumcísio Dómini nostri Jesu Christi, et Octáva Nativitátis ejúsdem.

[Circumcisio-Octava] (rubrica 196)
Octáva Nativitátis Dómini nostri Jesu Christi.
Circumcísio ejúsdem Dómini nostri Jesu Christi.

…
```

Then, the perl script in `specprima`only has to take care of adding the Luna to the Titulus and prepending the `r.` to the lines of Martyrologium and appending the `V. Et alibi ...`. Everything else is handled by `setupstring.pl`

FAJ-Munich:
[martyrologium-unmatched.csv](https://github.com/user-attachments/files/30790615/martyrologium-unmatched.csv)

@pythonengineer   Please find attached ~ 1500 hashes from vernaculars without a matching hash in Latin.

pythonengineer:
@FAJ-Munich

The reason why I didn't put an authoritative index of ordering in the Latin files is precisely because of the existence of the hundreds of vernaculars that aren't present in any of the Latins. So if we add an index it begs the question: where in the Latin order do we put a hash that only exists for example in English? Just according to how it already appears in the English?

Also, should we only render an elogium that is missing from the preferred language side according to the user's fallback language setting (as @mbab proposed, instead of just using the left-side)? And would we retain the new option for use of that or just assume so if a fallback is already set by the user?

FAJ-Munich:
The issue with the vernacular Martyrologium not matching the Latin is an issue indeed. However, part of developing this functionality should be to clear up the issues and mismatches. 

If, for instance, the Bohemice version has entries which are actually coming from the Cistercian Martyrology, then the files should be structured such that both, Latin and all vernaculars (not just Czech), show these Elogia for the Cistercian version (that would be @Augustinus-Altovadensis to help with).

If, in contrast, the issue just comes from mismatching the entries due to formatting, then it should be rectified, I guess. @APMarcello3 : Correct me if you think otherwise, but: Latin should always be the authorative database entry and the vernaculars should follow it, shouldn't they?
For English, for instance, the entries for January 2nd show a good example of a mismatch that can be resolved easily:
```
Latin:
[Tomis-Ponto]
Tomis, in Ponto, sanctórum fratrum Argéi, Narcíssi et Marcellíni púeri. Hic, sub Licínio Príncipe, cum inter tirónes esset comprehénsus et nollet militáre, hinc, cæsus ad mortem ac diu macerátus in cárcere, demum, in mare demérsus, martýrium consummávit; ejus autem fratres gládio perémpti sunt.

English:
[Tomis-Ponto]
At Tomi, in Pontus, under Emperor Licinius, the three holy brethren, Argeus, Narcissus, and Marcellinus. 

[Argeus-Narcissus]
Argeus and Narcissus were slain with the sword. Marcellinus was a boy, he was taken among the recruits, and for as much as he would not be a soldier he was grievously flogged, and after suffering long in prison was drowned in the sea (in the year 320.)
```

The mismatch comes from a misplaced linebreak in the current source files. The two English elogia should simply be one.

APMarcello3:
@FAJ-Munich Agreed.

pythonengineer:
> The issue with the vernacular Martyrologium not matching the Latin is an issue indeed. However, part of developing this functionality should be to clear up the issues and mismatches.
> 
> If, for instance, the Bohemice version has entries which are actually coming from the Cistercian Martyrology, then the files should be structured such that both, Latin and all vernaculars (not just Czech), show these Elogia for the Cistercian version (that would be @Augustinus-Altovadensis to help with).
> 

I see, so whatever entry is unique to a vernacular because of the Martyrology it uses should have its translation added into the Latin (which provides the schema) and then all the other languages? That sounds like quite the undertaking as far as getting full and synchronized coverage over all entries goes.

I'm wondering if a good start would be to first, just get the "keys" corresponding to those vernacular-unique elogia into the Latin index. Making sure that a value, i.e. translation actually exists for that elogium across all languages is a different story as that requires making or finding translations that don't currently exist in our files, but making sure the keys are now present at least so that text can be easily added to them incrementally improving coverage would be a starting point I could get moving on. That and of course resolving obvious mismatches.

Augustinus-Altovadensis:
@pythonengineer Yes, as @FAJ-Munich suggested, the entire Czech Martyrology is in fact the Cistercian Martyrology. I didn't have time to produce the Latin version, so what you see there is just Roman Latin Martyrology with a Czech translation of the Cistercian one. 

Also in his post from August 5, 3:57 PM, he very correctly suggests that there needs to be a file that will implement all the versions, including various differences in both Calendar and local traditions. The elogium for say St. Bernard will be very different in the Cistercian Martyrology when compared to Roman one, etc. 

I'll be happy to sort out the Czech as well as the Cistercian version, of course as soon as there is a Latin version to translate and a system that enables to work with various Martyrology versions. As far as I have understood, this hasn't been implemented yet (sorry, I don't currently have time to go through all the code).

But I agree that this has long been needed - with the modifications suggested above.

pythonengineer:
Several issues raised above have been implemented. An important note is that originally when this PR was posted my idea was to keep all rendering of days *exact* before and after changes to lessen the potential of "too many things being done in a single PR." Since there is a consensus that the vernacular should be generated from a "table" in the Latin file, now all elogia across all languages/versions are rendered according to this rule, which means that now there *are* display differences on many days. Primarily with ordering: if the English was listing elogia in a different order than the Latin, it's now arranged uniformly.

Issues raised above about the use of Python and folder placement are resolved. Tons of mismatching or vague keys have also been fixed. The new left-side fallback option was removed: it now uses the existing fallback language setting if an elogium is missing from the choice vernacular translation.

Nothing has been done to translate elogia missing from the Latin that exist in some vernacular, or elogia that are missing from a vernacular but present in the Latin martyrology. However, what has been done is that anything missing from the Latin has now had its name 'latinized' so that its key can be present in the table even if a value is not set, that way the organizational work is already done, uniform, and filling in a missing Latin elogium is as simple as putting in the text. This is necessary so that hundreds of elogia don't get suddenly removed from display just because it doesn't exist in the Latin.

Frankly, it needs some testing still to make sure everything is matched up and nothing's missing, but we're closer to an ideal system.

APMarcello3:
I am going to close this PR for now until sufficient work has been accomplished to make it ready for merge.

pythonengineer:
@APMarcello3 I don't know if reopening this PR will update this with the current state of commits on my branch, but I have pushed changes resolving the last few issues I could find, including a pre-existing problem with some languages like Spanish having English dates put at the top programmatically instead of the expected language, and the merge conflicts that were in the way.

The readme in the `admin/martyrologium` folder explains the syntax. The various issues discussed previously like mismatches and arbitrary key names are cleared up as far as I can see. The behavior as of a merge would entail default use of the user's fallback language where a translation is missing entries as demonstrated here in this big Czech example:

czech

The selected fallback language is used where our vernacular is missing an entry, as DO already does in other instances as discussed above. The exception is if the fallback elogium specifies a different rubric (other than the Roman) for its source martyrology. If use of the fallback we want toggleable, then that can easily be added, otherwise this should be good to merge.

PR #5438, opened 2026-08-12
Introitus -> Commune C10b

mpdc-p:
It is the same one, except for the Allelúja between parentheses

https://github.com/DivinumOfficium/divinum-officium/blob/master/web/www/horas/Latin/Commune/C10b.txt

### DIFF
--- a/web/www/missa/Latin/Sancti/08-14.txt
+++ b/web/www/missa/Latin/Sancti/08-14.txt
@@ -7,12 +7,7 @@ Prefatio=Communis
 Suffragium=Spiritu;;
 
 [Introitus]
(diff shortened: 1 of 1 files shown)


mpdc-p:
@APMarcello3 Hi Albert, I am new to this. Why hasn't this worked? If I put 08-14-2026 in the webpage, the Introito still shows the old version and not the referenced one in this update. I haven't uploaded the Spanish translation because it should be parametric from the referenced one.

APMarcello3:
@mpdc-p Please be patient as it takes time for changes to go live.

mpdc-p:
@APMarcello3 thanks, I thought I was doing something wrong

PR #5501, opened 2026-09-01
Repeated [PLEASE CHECK] + Translation + Uniformization

mpdc-p:
When using ex cX or vide cX, both take all prayers from that Commune reference, unless there is a specific prayer that day that overrides?, or it only works for vide? That's what needs to be checked.

### DIFF
--- a/web/www/horas/Latin/Commune/C10b.txt
+++ b/web/www/horas/Latin/Commune/C10b.txt
@@ -19,7 +19,7 @@ Vultum tuum (a Nativitate usque ad Purificationem)
 !Ps 44:13; 44:15-16
 v. Vultum tuum deprecabúntur omnes dívites plebis: adducéntur Regi Vírgines post eam: próximæ ejus adducéntur tibi in lætítia et exsultatióne. (Allelúja, allelúja.)
 !Ps 44:2
-Eructávit cor meum verbum bonum: dico ego ópera mea Regi.
+Eructávit cor meum verbum bonum: dico ego ópera mea regi.
 &Gloria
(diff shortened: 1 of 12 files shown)


web/www/missa/Espanol/Sancti/09-08.txt
FAJ-Munich: The Rank hash has to renamed Officium

web/www/missa/Latin/Sancti/07-02.txt
FAJ-Munich: I don't know by heart, but is C11 inheriting the Introitus and Commmunio from C10? Only then, they would be superfluous here.

web/www/missa/Latin/Sancti/09-08.txt
FAJ-Munich: See above

web/www/horas/Latin/Commune/C11.txt
rl1234567890, line 1: Missals have Regi in capital, since it refers to Christ. 
https://archive.org/details/sp07mr62-editio-typica/page/n853/mode/2up
Stet

web/www/horas/Latin/Commune/C10b.txt
rl1234567890, line 1: As for C11.txt, I would leave it alone, since  the King is Christ.

web/www/horas/Latin/Commune/C11.txt
mpdc-p: It is true, I was working on it from a Gradual, my bad.

mpdc-p:
@rl1234567890 thanks, corrected

PR #5522, opened 2026-09-09
Fixing some Neerlandia plus examples for further improvements

FAJ-Munich:
Dear @ntvangoor 

Thanks for your contributions and expansion of available dioceses. While the use of tools like Claude may certainly speed up a lot of processes, their output, unfortunately and typically, has to be taken with a grain of salt and requires human control and intervention. Without keeping the project's database as clean and correct as possible, we are at the risk of creating something unmaintable at some point. After all, _better is the enemy of the good._

Please have a look at the examples and explanations in this PR's commits and apply these principles to the other files for Utrecht and Groningen as well. In particular, I have noticed that the entries in _Tabulae_ do not include the commemorations due to occuring feasts of the General Calendar and that all provided feasts are activated for all rubrics (1570 – 1960) regardless of historicity. Without checking all the particular cases, I am fairly certain that some of the saints have not been canonised or have lived before 1570 (or 1888 etc.) even: to have masses or offices for them in those rubrics is less than ideal. For Utrecht, also, you provide Octave Days in the rank of Duplex majus for a number of feasts. For Tridentine rubrics, these have to be Duplex minus. Also, the (commemorations) of the dies infra 8vam are missing with their respective rank of Semiduplex. 

Once again, thanks for contributing and looking forward to the required updates.


ntvangoor:
@APMarcello3 I will certainly revisit my previous commits with that in mind. 
Thank you for the feedback!

PR #5524, opened 2026-09-10
Make webdia.pl resilient to blank rubric lines

jjh-servi:
Before this commit, there was a bug where a blank rubric line like `!  ` would cause a <font> tag to be emitted without any closing </font> tag, making all the following text in the section red.

For example, in the old Assumption Mass (rubrica 1939 or earlier), when displaying English, `missa.pl` outputs the introit without the citation `!Sedulius` and the entire introit text in red. `missa/English/Sancti/08-15t.txt` (old Assumption Mass) does not have give a proper `[Intoitus]`, so it falls back to the Latin `missa/Latin/Sancti/08-15t.txt`, which in turn pulls the introit from `@Sancti/07-16`. However, the English `web/www/missa/English/Sancti/07-16.txt` includes an exclamation mark followed by two spaces and a newline (`!  `), tripping up the `webdia.pl`, which output a `<font>` tag without a closing `</font>`.

### DIFF
--- a/web/cgi-bin/horas/webdia.pl
+++ b/web/cgi-bin/horas/webdia.pl
@@ -357,6 +357,7 @@ sub clean_setupsave {
 # returns <FONT ...>$text</FONT> string
 sub setfont {
   my $istr = shift;
+  my $tag_only = shift;
   my $text = shift;
   return $text unless $istr;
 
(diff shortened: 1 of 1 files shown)


jjh-servi:
Father Marcello, sorry for submitting a PR with code that broke `ordo.pl`. Unfortunately, this is beyond my Perl skills, so I opened issue #5529; hopefully somebody more knowledgeable can find a solution.

PR #5558, opened 2026-09-17
Runtime is Perl 5.40, as defined in the Dockerfile:15

JorgeRaimundo:
Updating the lint action to match the production runtime defined in the [Dockerfile:15](https://github.com/DivinumOfficium/divinum-officium/blob/62cffa75eb483c63030c206c0ad3c93a19137634/Dockerfile#L15)

Using dumb-init from the package manager.


JorgeRaimundo:
I pushed a commit ([7b7eae6](https://github.com/DivinumOfficium/divinum-officium/pull/5558/commits/7b7eae6ce220d4db5ecb48a42a85ef28d71e26cd)) after submitting this PR and it got incorporated here.
I wanted to do a separate PR for it.

APMarcello3:
@tjrandall will this cause any issues?

tjrandall:
> @tjrandall will this cause any issues?

@APMarcello3 - I built the PR branch and tested it. The image builds cleanly, dumb-init is the same 1.2.5 version we were downloading before (now from the Debian package at /usr/bin/dumb-init), and the container starts under it and serves the site normally. Nothing in the repo references the old path, and our Cloud Run service doesn't override the entrypoint, so the deploy will pick up the new path automatically. The lint change moves CI from Perl 5.28 to 5.40 to match production. That workflow only runs on PRs that touch .pl or .pm files, so it won't run on this one, but I ran the same perltidy check locally under Perl 5.40 against the whole repo and it came back clean. This is fine to merge.

